PR #10, opened 2025-10-18
Fix account handling and refactor authentication

mohit:
This PR contains two commits that fix bugs and improve the authentication system:

## Commit 1: Fix AttributeError in account handling
- Fixes `AttributeError: 'NoneType' object has no attribute 'get'` when `institution` or `type` fields are None
- Changes `account.get('institution', {})` to `(account.get('institution') or {})`
- Changes `account.get('type', {})` to `(account.get('type') or {})`
- Prevents crashes in `get_accounts()` when account data has None values

## Commit 2: Refactor authentication to remove global state
- Removes global `_monarch_client` variable from server.py
- Removes `self.mm` instance variable from `SecureMonarchSession`
- `get_monarch_client()` now creates fresh client each call from keyring token
- `get_authenticated_client()` returns new MonarchMoney instance instead of cached
- Fixes import path to use absolute path: `monarch_mcp_server.secure_session`
- Adds `AccessTokenT` import for future auth improvements
- Improves statelessness and reduces potential auth bugs from stale cached clients

Both changes improve robustness and fix real-world errors encountered during usage.

### DIFF
--- a/src/monarch_mcp_server/secure_session.py
+++ b/src/monarch_mcp_server/secure_session.py
@@ -14,25 +14,23 @@ logger = logging.getLogger(__name__)
 KEYRING_SERVICE = "com.mcp.monarch-mcp-server"
 KEYRING_USERNAME = "monarch-token"
 
+
 class SecureMonarchSession:
     """Manages Monarch Money sessions securely using the system keyring."""
-    
-    def __init__(self):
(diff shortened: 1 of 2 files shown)


mohit:
I was running into an issue with `NoneType` for some of my accounts and losing auth if there was an error in one chat in others.

PR #20, opened 2026-02-08
Add transaction split support

yonib05:
This adds support for splitting transactions into multiple parts with different categories or merchants.

## Changes

Added two new tools:

- `get_transaction_splits` - View how a transaction has been divided into parts
- `split_transaction` - Split a single transaction into multiple parts with different categories or merchants

## Use Case

Split transactions are useful when a single purchase contains items from multiple budget categories. For example, splitting a Costco trip into groceries and household items, or separating a mixed Amazon order into different expense categories.

## Implementation

Uses the existing `get_transaction_splits` and `update_transaction_splits` methods from the monarchmoney library. The split amounts must sum to the original transaction amount.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,12 @@ Once authenticated, use these tools directly in Claude Desktop:
 - **Delete Transaction**: Remove a transaction
 - **Get Recurring Transactions**: View upcoming recurring transactions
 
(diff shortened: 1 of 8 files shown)


robcerda:

PR #21, opened 2026-02-08
Add bulk categorization support

yonib05:
## Summary

Adds the ability to categorize multiple transactions at once.

## What's New

- `bulk_categorize_transactions` - Apply a category to multiple transactions in a single call

Accepts a list of transaction IDs and a category ID. Returns success/failure status for each transaction.

## Why

When you have multiple similar transactions (like recurring subscriptions or multiple purchases from the same merchant), it's tedious to categorize them one by one. This lets you select them all and categorize in one go.

## Testing

Added tests for bulk operations including partial failure scenarios.
Tested with batches of 5-10 transactions on my account.


robcerda:

PR #27, opened 2026-02-08
Add transaction rules API support

yonib05:
## Summary

Adds tools for managing automatic transaction categorization rules.

## What's New

- `get_transaction_rules` - Lists all configured rules
- `create_transaction_rule` - Creates a new rule with conditions and actions
- `update_transaction_rule` - Modifies an existing rule
- `delete_transaction_rule` - Removes a rule

Supports:
- Merchant name matching (equals, contains, starts with, ends with)
- Amount conditions (greater than, less than, equals, between)
- Setting category, merchant name, tags
- Hiding from reports

## Technical Note

The rules API is not part of the official monarchmoney Python library. I reverse-engineered the GraphQL mutations by inspecting network traffic from the Monarch web app. The implementation uses `gql_call` directly with the discovered query structures.

This should be considered somewhat experimental - if Monarch changes their internal API, these tools may break. But the rules functionality is too useful to leave out.

## Testing

New test file `tests/test_rules.py` with mocked API responses.
Tested rule creation and deletion on live account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,12 @@ Once authenticated, use these tools directly in Claude Desktop:
 - **Delete Transaction**: Remove a transaction
 - **Get Recurring Transactions**: View upcoming recurring transactions
 
+### 🤖 Transaction Rules (Auto-Categorization)
+- **Get Transaction Rules**: List all auto-categorization rules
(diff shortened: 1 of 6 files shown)


robcerda:

PR #28, opened 2026-02-08
Add Google OAuth authentication support

yonib05:
## Summary

Adds support for users who sign in to Monarch with their Google account.

## What's New

- `google_login.py` - Standalone script for browser-based Google OAuth login
- `authenticate_with_google` - MCP tool that triggers the OAuth flow
- Updated `setup_authentication` tool to mention both auth options
- Added `playwright` dependency for browser automation

## How it works

1. Opens a real browser window (Chromium via Playwright)
2. Navigates to Monarch login page
3. User completes Google OAuth in the browser
4. Script intercepts the auth token from API requests
5. Token is saved to keyring for future use

## Why

The existing `login_setup.py` only works with email/password auth. Many users (myself included) use Google Sign-In and don't have a Monarch password. This was blocking adoption for those users.

## Security

- No credentials are stored by the script
- Token is captured from legitimate browser session
- Stored in system keyring (same as existing auth)
- Browser runs in non-headless mode so user can verify the login page

## Testing

Tested with my own Google-linked Monarch account.
Token capture works reliably after OAuth redirect.

## Note

Requires `playwright` package and browser install (`playwright install chromium`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,12 @@ Once authenticated, use these tools directly in Claude Desktop:
 - **Delete Transaction**: Remove a transaction
 - **Get Recurring Transactions**: View upcoming recurring transactions
 
+### 🤖 Transaction Rules (Auto-Categorization)
+- **Get Transaction Rules**: List all auto-categorization rules
+- **Create Transaction Rule**: Create rules with merchant/amount conditions to auto-categorize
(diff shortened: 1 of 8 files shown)


robcerda:

PR #45, opened 2026-04-18
Replace login_setup.py with MCP elicitation-based login

### LINKED ISSUE #44: Replace login_setup.py with MCP elicitation-based login
**Summary**                                                   

The current auth flow requires the user to run python login_setup.py in a separate terminal, inside the project directory, whenever a session is needed or expires. This is awkward for end users — especially in Claude Code, where the path and cwd aren't always obvious — and the 190-line login_setup.py duplicates logic the MCP server already has. It also hardcodes "Claude Desktop" in user-facing strings even when invoked from Claude Code or another client.
                                                                                                                                                                                                                                       
MCP now supports elicitation, which is designed for exactly this: a tool can ask the client to prompt the user for structured input mid execution, without the values ever appearing as tool arguments in the model's context.       

**Proposal**                                                                                                                                                                                                                             
                                                            
  Replace login_setup.py with in-server tools:                                                                                                                                                                                         
   
  - monarch_login() — calls ctx.elicit(...) with a Pydantic schema containing email: str, password: str (format="password"), and an optional mfa_code: str. The client UI (Claude Desktop, Claude Code, etc.) renders the form,        
  collects values, and returns them to the server over the protocol. If Monarch raises RequireMFAException and mfa_code was blank, re-elicit for just the MFA code and retry multi_factor_authenticate. On success, persist the session
   token via the existing secure_session keyring path.                                                                                                                                                                                 
  - monarch_login_with_token(token: str) — preserved for SSO users who paste a browser-localStorage token.
  - monarch_logout() — clears the keyring entry.                                                                                                                                                                                       
  - Delete login_setup.py, setup_authentication, and debug_session_loading.                                                                                                                                                            
                                                                                                                                                                                                                                       
  **Why this is safe**                                                                                                                                                                                                                     
                                                                                                                                                                                                                                       
  MCP elicitation has two modes (see mcp.server.fastmcp.Context):                                                                                                                                                                      
                                                            
  - Form mode (ctx.elicit) — values flow client-UI → server directly via a protocol message. They do not pass through the model as tool arguments, are not written to the conversation transcript, and the model cannot see them.      
  Suitable when the MCP client itself is trusted (typical for local stdio servers like this one).
  - URL mode (ctx.elicit_url) — for higher-security flows (OAuth, payments) where the credential must not touch the MCP client either. Not used here because it requires the server to host an out-of-band URL.                        
                                                                                                                                                                                                                                       
  For a user-trusted stdio server talking to a password-based API, form mode is the right fit and strictly better than the current design — credentials currently pass through an interactive shell prompt and a keyring write         
  initiated by a separately-run Python process, which is both less convenient and not materially more secure.                                                                                                                          
                                                                                                                                                                                                                                       
  **Rough sketch**                                              

  from pydantic import BaseModel, Field

  class LoginForm(BaseModel):                                                                                                                                                                                                          
      email: str
      password: str = Field(..., json_schema_extra={"format": "password"})                                                                                                                                                             
      mfa_code: str | None = None                           
                                                                                                                                                                                                                                       
  @mcp.tool()
  async def monarch_login(ctx: Context) -> str:                                                                                                                                                                                        
      result = await ctx.elicit("Sign in to Monarch Money", LoginForm)
      if result.action != "accept":                                                                                                                                                                                                    
          return "Login cancelled."                                                                                                                                                                                                    
      form = result.data                                                                                                                                                                                                               
                                                                                                                                                                                                                                       
      mm = MonarchMoney()                                                                                                                                                                                                              
      try:
          await mm.login(form.email, form.password, use_saved_session=False, save_session=False)                                                                                                                                       
      except RequireMFAException:                                                                                                                                                                                                      
          code = form.mfa_code
          if not code:                                                                                                                                                                                                                 
              mfa_result = await ctx.elicit("Enter MFA code", MFAForm)
              if mfa_result.action != "accept":                                                                                                                                                                                        
                  return "Login cancelled."                                                                                                                                                                                            
              code = mfa_result.data.mfa_code
          await mm.multi_factor_authenticate(form.email, form.password, code)                                                                                                                                                          
                                                                                                                                                                                                                                       
      secure_session.save_authenticated_session(mm)                                                                                                                                                                                    
      return "Logged in. Session saved."                                                                                                                                                                                               
                                                                                                                                                                                                                                       
  **Benefits**                                                  

  - No separate script, no cd into the project directory.                                                                                                                                                                              
  - Re-auth is one sentence in chat when a session expires.
  - Client-agnostic — no more Claude-Desktop-specific copy.                                                                                                                                                                            
  - ~150 lines of code deleted net.                                                                                                                                                                                                    
                                                                                                                                                                                                                                       
  **Caveat**                                                                                                                                                                                                                               
                                                                                                                                                                                                                                       
  Requires an MCP client that implements elicitation. Claude Code and current Claude Desktop both do. Older clients can still use the token-paste path (monarch_login_with_token) as a fallback.                                       
   
  Happy to send a PR if you'd like.

darrenhaas:
Closes #44.

## Summary

Replaces the standalone `login_setup.py` script with in-server tools that use [MCP elicitation](https://modelcontextprotocol.io) (form mode) to collect credentials. Net: **+441 / −923 lines**.

**New tools:**
- `monarch_login` — password + MFA sign-in via an elicitation form
- `monarch_login_with_token` — paste a browser session token (SSO users)
- `monarch_logout` — clear the stored session
- `check_auth_status` — rewritten to actually verify the stored token is live (was previously only checking for presence in the keyring)

**Removed:**
- `login_setup.py` (190-line standalone script)
- `setup_authentication` tool (stale instructional text)
- `debug_session_loading` tool (ad-hoc debug output)
- `MONARCH_EMAIL` / `MONARCH_PASSWORD` env var bypass in `get_monarch_client` (authentication is now always explicit)

**Non-breaking:**
- All data-tool signatures (`get_accounts`, `get_transactions`, etc.) unchanged.
- `secure_session.py` keyring/file-fallback layer unchanged — existing saved sessions keep working.

## Why

The current flow requires the user to:

1. Open a terminal.
2. `cd` into the checkout directory.
3. Run `python login_setup.py` (or `uv run python login_setup.py`).
4. Re-enter credentials every time the session expires.

Inside Claude Code this is especially awkward — the cwd and venv aren't always obvious — and the 190-line script duplicates logic the MCP server already has. It also hardcodes "Claude Desktop" in user-facing strings even when invoked from other clients.

MCP elicitation is designed for exactly this: a tool can ask the client to prompt the user for structured input mid-execution.

## Security

The `mcp.server.fastmcp.Context` class exposes two elicitation modes:

- **Form mode** (`ctx.elicit`) — values flow client-UI → server directly via a protocol message. They do **not** appear as tool arguments, are not written to the conversation transcript, and the model cannot see them. Used here.
- **URL mode** (`ctx.elicit_url`) — for higher-security flows where the credential must not touch the MCP client either (OAuth, payments). Not used here because it would require the server to host an out-of-band URL.

For a user-trusted stdio server talking to a password-based API, form mode is a strict upgrade over the status quo — credentials currently pass through an interactive shell prompt and a keyring write initiated by a *separately-run* Python process, which is neither more convenient nor materially more secure.

The `monarch_login` tool exposes **zero arguments** in its input schema:

```
monarch_login → {"properties":{},"title":"monarch_loginArguments","type":"object"}
```

Credentials can only enter the server via elicitation.

## Schema compatibility with Claude Code

Claude Code's MCP client validates the `requestedSchema` with a strict Zod validator (verified against 2.1.114 via session transcript). Two forms of the schema that Pydantic generates by default will be **rejected**:

- `json_schema_extra={"format": "password"}` on a string field. Only `"email"`, `"uri"`, `"date"`, and `"date-time"` are accepted as `format` values. There is no password-masking hint in the MCP elicitation spec — the client chooses its UX.
- `Optional[str]` / `anyOf`. Every property must declare a single concrete type. Null is not allowed. Optional fields must be modeled as a separate, conditional follow-up elicitation.

This PR's `LoginForm`, `MFAForm`, and `TokenForm` comply with both constraints. MFA is collected via a second elicitation call if and only if Monarch raises `RequireMFAException`.

## Other cleanup in this PR

- Drop the `run_async(ThreadPoolExecutor)` dance in every tool; use async tools natively. FastMCP awaits async tool functions directly.
- Remove "Claude Desktop"-specific wording from tool docstrings and return values.
- Rewrite the README around the new flow and clean up the tool table.

## User-visible migration

Users upgrading just need to call `monarch_login` once in chat. Existing keyring sessions continue to work — no forced re-auth.

## Caveat

Requires an MCP client that implements elicitation. Claude Code (2.1.76+) and current Claude Desktop both do; older clients can fall back to `monarch_login_with_token` (token paste).

## Test plan

- [x] Restart the MCP server, call `monarch_login` → form appears in client UI → enter creds (including MFA) → returns "Logged in. Session saved to system keyring." *(verified on Claude Code 2.1.114)*
- [ ] Call `get_accounts` → returns account list
- [ ] Call `check_auth_status` → "Authenticated — session is live."
- [ ] Call `monarch_logout` → "Cleared stored Monarch session."
- [ ] Call `get_accounts` again → returns "Not signed in to Monarch Money. Call `monarch_login`."
- [ ] Test SSO flow via `monarch_login_with_token`.


darrenhaas:
Opened in error — this should have stayed in my fork while I test it locally. Reopening later if/when ready for upstream review.

robcerda (dismissed):
Thanks for this — the MCP elicitation pattern is a real security improvement (credentials never hit the model's context), and dropping the separate \`login_setup.py\` flow is a nice cleanup.

However, this PR breaks all 53 existing tests in \`tests/test_tools.py\` and \`tests/conftest.py\`:

- \`conftest.py\` patches \`monarch_mcp_server.server.get_monarch_client\`, which this PR renames to \`_get_client\`
- Tool functions are now \`async\` rather than sync-wrapping \`run_async\` — the existing tests call them synchronously
- The try/except blocks were removed from every tool, so tests asserting on error strings like \`"Error getting accounts"\` will fail

Could you update the tests to match the new shape? A few pointers:

- Patch \`monarch_mcp_server.server._get_client\` (sync, no \`AsyncMock\`) returning the mock MonarchMoney client
- Wrap test calls with \`asyncio.run(...)\` or mark the tests \`@pytest.mark.asyncio\`
- For the removed error-handling: either drop those test cases, or (my preference) add the try/except back at the tool level so a tool raising an exception doesn't crash the MCP request

Running \`uv run --extra dev pytest tests/\` should get you to green. Happy to review again once tests pass.

robcerda:
@darrenhaas ,  I have a working implementation of #44 ready locally (elicitation-based login tools with tests). Since `maintainerCanModify` is on, I could push it to this branch, or I can open a separate PR, whichever you prefer. Happy to hold if you'd rather finish your own first. Just let me know what works best.

darrenhaas:
Perfect, thank you Rob.

PR #49, opened 2026-05-01
Improve Monarch transaction search metadata

mateobelanger:
Summary:
- add get_transactions response envelope metadata for spooled agent-tools output
- expose original statement, currency, direction metadata, category group data, and merchant id
- add wide_search fallback for empty/erroring Monarch search results across original statement, merchant, notes, category, account, and tags
- add category_group_ids filtering

Tests:
- uv run pytest -q

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 | `setup_authentication` | Get setup instructions | None |
 | `check_auth_status` | Check authentication status | None |
 | `get_accounts` | Get all financial accounts | None |
-| `get_transactions` | Get transactions with filtering | `limit`, `offset`, `start_date`, `end_date`, `account_id` |
+| `get_transactions` | Get transactions with filtering and reconciliation fields | `limit`, `offset`, `start_date`, `end_date`, `account_id`, `account_ids`, `category_ids`, `category_group_ids`, `tag_ids`, `search`, `wide_search`, `search_scan_limit`, `has_notes`, `is_split`, `is_recurring` |
 | `get_budgets` | Get budget information | `start_date`, `end_date` |
 | `set_budget_amount` | Set budget for a category | `amount`, `category_id`, `category_group_id`, `start_date`, `apply_to_future` |
 | `get_cashflow` | Get cashflow analysis | `start_date`, `end_date` |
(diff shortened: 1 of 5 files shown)


robcerda:
Thanks for the PR. The new fields (`original_statement`, `plaid_description`, `category_group_ids` filter) are useful. A few things to address before merging:

**Reuse existing helpers.** `helpers.py` already has `json_error()` (same shape as your inline envelope) and `format_transaction()`. `_format_exception()` is a real improvement but should land in `helpers.py` so all tools pick it up. Same for `_format_transaction_row`: extend `format_transaction()` rather than running a parallel formatter.

**Lazy load category metadata.** `_load_category_metadata()` currently runs on every `get_transactions()` call, adding a round trip even when the caller does not need `category_group` fields. Only fetch when `category_group_ids` is set or wide_search needs it.

**Breaking response shape.** Existing callers expect a JSON list. The new envelope is useful but silently breaks any agent doing `transactions[0]`. Either gate it behind a flag (`envelope=False` default) or flag it as a breaking change in the PR description so I can bump the version and update the CHANGELOG.

**`wide_search=True` default.** With `search_scan_limit=1000`, every empty server side search pulls 1000 rows for local filtering. Reasonable as opt in, heavy as default. Suggest defaulting to `False`, or dropping the scan limit to 200.

**Currency heuristic.** Regex matching account display names reports `currency: "USD"` for "Wise USD Balance" while returning `null` for "Fidelity Brokerage". The LLM will read the guess as fact. Either drop it or add a `currency_source` field so the consumer knows when the value is heuristic.

**Retry chain.** The server then lowercase then wide search nesting is hard to follow. Most search backends are already case insensitive, so I would drop the lowercase retry and keep just server then wide search.

Happy to look at a revision. Tests are clean and the security read is fine, this is purely about fitting existing patterns.

robcerda:
@mateobelanger I went ahead and pushed follow-up commits to your branch via \`maintainerCanModify\` so we can keep momentum without you having to do the legwork. Your original commit is preserved at the bottom of the stack with your authorship intact. Five commits on top from me:

1. **Move reusable helpers into helpers.py** — \`format_exception\`, \`first_present\`, \`tool_response_envelope\`. \`json_error\` now uses \`format_exception\` so all tools benefit from the empty-message fallback.
2. **Lazy-load category metadata** — only fetched when \`category_group_ids\` is set or when \`wide_search\` runs. Ordinary calls drop the extra round trip and rely on the inline \`category.group\` Monarch already returns.
3. **\`wide_search\` opt-in by default** — flipped to \`False\` and lowered \`search_scan_limit\` to 200. Empty server-side searches no longer pull 1000 rows automatically.
4. **\`currency_source\` label** — kept the regex heuristic but the response now includes \`currency_source: "api" | "account_name_guess" | null\` so consumers know when the value is a guess.
5. **Simplified retry chain** — dropped the lowercase casefold retry. New flow: server search, optional \`wide_search\` on empty/error, otherwise propagate. 60 lines lighter.

All 139 tests pass.

**One thing I left as-is and want your read on**: the response envelope is still the default (was a JSON list, now \`{tool, args, count, total_count, truncated, search, data}\`). It is technically a breaking change for any caller doing \`transactions[0]\`, but the envelope adds real signal. I am leaning take the break + bump the project to 0.2.0 in pyproject.toml when this lands, but I am open to gating it behind \`envelope=True\` if you would rather keep backward compatibility. What is your preference?

Thanks again for the original PR. The new fields and \`category_group_ids\` filter are good wins.

PR #53, opened 2026-05-08
Add splits_info support, return rule_id on create, normalize formatter contract

micahldixon:
## Summary

Closes four gaps in `src/monarch_mcp_server/tools/rules.py` that prevent the MCP tool from being used for split-rule operations.

1. **`create_transaction_rule` / `update_transaction_rule` now accept `splits_info`.** Maps a list-of-dicts `splits_info` parameter (snake_case keys) to the camelCase shape Monarch expects, so split rules can be created without dropping to raw GraphQL.

2. **`create_transaction_rule` returns the new `rule_id`.** Re-fetches `transactionRules` and finds the new rule by matching every defining criterion. Callers previously had no way to address the rule they just created.

3. **`get_transaction_rules` formatter always emits action keys.** Always emits the keys with `None` values when null so downstream callers can rely on key presence.

4. **Removed dead "wait + retry" guidance.** `applyToExistingTransactions=true` is empirically synchronous — every retroactive sweep observed in testing returned populated splits on the very first `getTransactionDetails` call.

## Test plan

- [x] Smoke-tested live: created a 2-line split rule via the new `splits_info` parameter, asserted `success=True` and presence of `rule_id`, then deleted the test rule.
- [x] Asserted `split_transactions_action` is non-null for a known split rule and `None` (key present) for a non-split rule.
- [x] `grep -nE "wait.*retry|retry.*wait|sleep.*[0-9]"` returns zero matches in `src/`.

### DIFF
--- a/src/monarch_mcp_server/tools/rules.py
+++ b/src/monarch_mcp_server/tools/rules.py
@@ -84,6 +84,22 @@ query GetTransactionRules {
     }
     setHideFromReportsAction
     reviewStatusAction
+    sendNotificationAction
+    splitTransactionsAction {
+      amountType
+      splitsInfo {
+        amount
+        categoryId
+        merchantName
+        tags
+        hideFromReports
+        needsReview
+        reviewStatus
+        goalId
+        __typename
+      }
+      __typename
+    }
     recentApplicationCount
     lastAppliedAt
     __typename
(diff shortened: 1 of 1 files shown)


micahldixon:
Closing — submitted in error.

PR #56, opened 2026-05-14
Add merchant management and recurring stream review tools

asreynolds1000:
## Summary

- **`get_merchant`**: Query a merchant's details including recurring transaction stream configuration (frequency, amount, base date, active status). Needed to inspect current state before updates.
- **`update_merchant`**: Modify a merchant's name and/or recurring stream settings. This is the only way to fix recurring stream configurations — there is no direct "update recurring stream" endpoint in the Monarch API.
- **`review_recurring_stream`**: Accept, ignore, or reset recurring transaction streams that Monarch detects automatically.

All three tools use custom GraphQL via `gql_call`, following the same pattern as the existing transaction rules tools in `tools/rules.py`.

## Motivation

During a financial review session, we needed to fix recurring stream configurations (frequency, amount, base date) for several merchants. The only way to do this was through raw GraphQL calls because the MCP server had no merchant management tools. These three tools cover the gap.

## Changes

- New module: `src/monarch_mcp_server/tools/merchants.py` (3 tools, 3 GraphQL operations)
- New test file: `tests/test_merchants.py` (14 tests)
- Wiring: `tools/__init__.py`, `server.py` re-exports, `conftest.py` `_TOOL_MODULES`
- README: new feature section, tool table entries, usage examples, mutating tools list

## Design decisions

- **No client-side validation of `frequency` or `review_status` values** — documented known values in docstrings but let the API validate, so the tools don't reject valid future values.
- **`update_merchant` validates that at least one field is provided** — prevents pointless no-op API calls.
- **GraphQL errors passed through unchanged** — follows the `rules.py` pattern of returning `{"success": false, "errors": <raw API errors>}` rather than parsing the error structure.
- **`get_merchant` wraps result in `{"merchant": ...}`** — both success and not-found paths share the same top-level key for consistent return shape.

## Test plan

- [x] `uv run pytest tests/test_merchants.py -v` — 14/14 passed
- [x] `uv run pytest tests/ -v` — 153/153 passed (0 regressions)
- [x] `black` and `isort` formatting applied
- [x] Wiring audit: verified all 4 registration points (\_\_init\_\_.py, server.py, conftest.py, README)

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,11 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 - **Update Transaction Rule**: Modify existing rules
 - **Delete Transaction Rule**: Remove a rule
 
+### 🔄 Merchant & Recurring Stream Management
+- **Get Merchant**: View a merchant's details including recurring transaction stream configuration
+- **Update Merchant**: Modify a merchant's name and/or recurring stream settings (frequency, amount, base date)
+- **Review Recurring Stream**: Accept, ignore, or reset recurring transaction streams detected by Monarch
+
 ### ✂️ Transaction Splits
 - **Get Transaction Splits**: View how a transaction has been split into parts
 - **Split Transaction**: Divide a single transaction into multiple parts with different categories or merchants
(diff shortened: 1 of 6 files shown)


fr3nch13:
This would be super helpful for me as well.

PR #59, opened 2026-05-17
Add cookie-based auth for Monarch's May 2026 API change

mattietea:
- **New `MonarchMoneyCookieAuth`** (`src/monarch_mcp_server/cookie_auth.py`) — subclass of `MonarchMoney` that overrides `_get_graphql_client()` to send cookies via `AIOHTTPTransport` with the `x-csrftoken` header.
- **`secure_session.py`** — added `save_cookies` / `load_cookies` / `delete_cookies` keyring helpers. `get_authenticated_client()` now prefers cookies and falls back to the legacy token. `save_authenticated_session()` dispatches on `isinstance(MonarchMoneyCookieAuth)`.
- **MCP elicitation** — added `CookiesForm` schema, `login_with_cookies_interactive()`, and the `monarch_login_with_cookies` tool. `check_auth_status()` now warns when only a legacy token is stored. `setup_authentication()` reflects the API change.
- **`login_setup.py`** — reordered the terminal menu so session cookies are the default option; moved the MFA-enabled prompt to only fire for the email/password path.

### DIFF
--- a/login_setup.py
+++ b/login_setup.py
@@ -19,6 +19,7 @@ sys.path.insert(0, str(src_path))
 
 from monarchmoney import MonarchMoney, RequireMFAException
 from dotenv import load_dotenv
+from monarch_mcp_server.cookie_auth import MonarchMoneyCookieAuth
 from monarch_mcp_server.secure_session import secure_session
 
 async def main():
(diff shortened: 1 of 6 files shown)


robcerda:
Security review before this leaves draft:

1. Combined secret storage. session_id and csrftoken live in one keyring entry and one JSON file. Disclosure gives both halves at once. Consider separate entries.

2. TOCTOU on cookie file. _save_cookies_file writes with default umask (usually 0644) then chmods 600. Fix with os.open(path, O_CREAT|O_EXCL|O_WRONLY, 0o600). Legacy token file has the same flaw.

3. Private attribute access. secure_session reads mm._session_id and mm._csrftoken. Add a public accessor on MonarchMoneyCookieAuth.

4. No upper bound on monarchmoneycommunity. The subclass touches private internals (_headers, _timeout, _get_graphql_client). pyproject is >=1.2.0 with no ceiling, so a point release can break auth silently. Issue #60 was exactly this failure mode.

5. Upload paths fail under cookie auth. _upload_form_data still uses the legacy Authorization header, so upload_account_balance_history and any future attachment tool will silently fail. Either guard at the tool boundary or note it in the docstring.

6. CookiesForm uses plain str fields. Pydantic SecretStr would let MCP clients mask input.

7. Test gaps. No file fallback coverage, no priority test (cookies preferred when both cookies and a legacy token are stored).

Solid baseline. These are tightenings, not blockers.

robcerda:
Closing as stale and superseded by the direction in #71, which uses Monarch's primary token plus device UUID auth path with email OTP support. Reopen if that direction stalls and the cookie approach is needed as a fallback.

PR #61, opened 2026-05-20
Default to read-only; disable credential/auth mutations

sakhalsa:
## Summary

Hardens the MCP server so that, by default, it cannot mutate remote Monarch state or change locally stored auth.

- **Read-only default.** New `MONARCH_MCP_READ_ONLY` env flag (defaults on). Every Monarch SDK mutation tool now refuses to run while read-only is active: `create_transaction`, `update_transaction`, `categorize_transaction`, `update_transaction_notes`, `mark_transaction_reviewed`, `bulk_categorize_transactions` (real run; dry-run still allowed), `delete_transaction`, `set_transaction_tags`, `create_transaction_tag`, `add_transaction_tag`, `create_transaction_category`, `split_transaction`, `create_transaction_rule`, `update_transaction_rule`, `delete_transaction_rule`, `set_budget_amount`, `refresh_accounts`, and `upload_account_balance_history` (real run; dry-run still allowed). Operators can flip the flag to a falsey value (`false`/`0`/`no`/`off`) to permit mutations explicitly.
- **Auth mutation tools hard-disabled.** `monarch_login`, `monarch_login_with_token`, and `monarch_logout` no longer accept credentials or modify stored auth state — regardless of the read-only flag. Credentials never flow through MCP transport; a remote MCP client cannot wipe the stored session. Authentication must be performed out-of-band via `python login_setup.py`.
- **No env credential auto-login.** Removed the `MONARCH_EMAIL` / `MONARCH_PASSWORD` fallback from `get_monarch_client`, removed `load_dotenv()` from `app.py` and `login_setup.py`, and dropped `python-dotenv` from the dependency list. Passwords are never silently read from the environment.
- **Read-only tools unaffected.** Account, transaction, summary, financial, category-listing, and tag-listing read tools continue to function as before.

## Testing

- `python -m pytest` — **181 passed**.
- New `tests/test_read_only.py` (39 tests) proves every mutation tool refuses in read-only mode, that the upstream MonarchMoney client is never invoked when a refusal fires, and that dry-run paths remain usable.
- Updated `tests/test_auth.py` (17 tests) proves the MCP auth-mutation tools are hard-disabled regardless of the read-only flag, that `MONARCH_EMAIL` does not trigger auto-login, and that neither `client.py` nor `app.py` reintroduce `os.getenv` of credential vars or `load_dotenv`.
- `tests/conftest.py` adds an autouse fixture that opts existing mutation-test suites out of read-only mode so they continue to exercise the underlying client wiring.

## Remaining risks

- The standalone `login_setup.py` terminal script still performs login; it is now the only path that writes to the keyring. That's by design but means operator hygiene around running that script matters.
- `dry_run=True` paths for `bulk_categorize_transactions` and `upload_account_balance_history` are intentionally kept open in read-only mode. They do not call the mutation SDK methods, but they DO call read endpoints (e.g. fetching balance history). Reviewers may want to confirm that's acceptable.
- README still documents the previous elicitation login flow; this PR does not rewrite the README. A follow-up doc pass is recommended.

---
🤖 *Generated by Computer*

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 A Model Context Protocol (MCP) server for integrating with the Monarch Money personal finance platform. This server provides seamless access to your financial accounts, transactions, budgets, and analytics through Claude Desktop and Claude Code.
 
+> **🔒 Hardened build (read-only by default).** As of the read-only hardening change, the server refuses all Monarch data mutations unless the operator explicitly opts in, and the MCP-exposed login/logout tools are permanently disabled. See [🔒 Security model](#-security-model) before configuring a client.
+
 My MonarchMoney referral: https://www.monarchmoney.com/referral/ufmn0r83yf?r_source=share
 
 **Built with the [MonarchMoneyCommunity Python library](https://github.com/bradleyseanf/monarchmoneycommunity)** - An actively maintained community fork of the Monarch Money API with full MFA support.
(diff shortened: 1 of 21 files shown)


robcerda:
The fail safe default and the string grep regression guards against load_dotenv and os.getenv MONARCH_EMAIL are excellent. A few items before this leaves draft:

1. Block list is already stale. PRs #56 and #62 just landed and added update_merchant, review_recurring_stream, and update_category. None are in the read only list. Structural fix: a @mutating decorator that auto registers, so a new mutating tool cannot be added without being guarded.

2. cookie_auth.py is duplicated with #59. Whichever lands second needs rebase. Suggest landing #59 first in isolation, then layering hardening on top here.

3. Cookie file TOCTOU. Same flaw as #59 (write_text then chmod 600). Fix with os.open(path, O_CREAT|O_EXCL|O_WRONLY, 0o600).

4. MONARCH_PASSWORD not surfaced in check_auth_status. Only MONARCH_EMAIL is reported as unused. A user could leave PASSWORD in their MCP config plaintext indefinitely with no warning.

5. PR scope. Three concerns bundled (read only default, auth tools hard disabled, env credential and dotenv removal). Each is independently shippable and bisectable, worth splitting.

6. refresh_accounts argument bug landed separately in #63. No action needed here, just flagging the order dependency.

The shape of the guard is right.

robcerda:
Agreed on the goal, but defaulting to read only and reworking the auth surface would break every existing install. Landed the opt in half as MONARCH_MCP_READ_ONLY in #129, which skips registering the mutating tools entirely, so closing this.

PR #68, opened 2026-06-04
Adds 'current_balance' and 'display_balance' to accounts output.

fr3nch13:
Monarch has two ways to show balances for an account's balance, for example, a checking account.

This PR adds `current_balance` and `display_balance` to the return of the account's details and keeps the  `balance ` in place for backwards compatibility. It also matches the key names with the monarch api, and makes it easier for LLMs to understand the values through inference of the name.


fr3nch13:
Can you merge this PR please?

PR #69, opened 2026-06-09
Add save_token.py for non-interactive token-based authentication

tyjud:
## Summary

- Adds save_token.py, a lightweight CLI script that accepts a Monarch Money session token as a command-line argument and saves it to the system keyring
- Bypasses the getpass-based interactive prompts in login_setup.py that block paste in many terminal emulators
- Targets users who authenticate via Google/SSO and need to copy their token from the browser

## Motivation

login_setup.py uses getpass.getpass() for credential input, which disables paste in many terminal environments. Users signing in with Google/SSO have no fallback since they cannot type a long session token by hand. save_token.py sidesteps this entirely by accepting the token as a positional CLI argument so it can be pasted into the command before pressing enter.

## Usage

uv run python save_token.py YOUR_TOKEN_HERE

## Test plan
- Copy session token from app.monarchmoney.com Local Storage
- Run: uv run python save_token.py <token>
- Confirm Connected message with account count
- Confirm MCP get_accounts tool works after saving

### DIFF
--- a/save_token.py
+++ b/save_token.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python3
+"""Quick token-based login — pass your token as a command-line argument."""
+import asyncio, sys
+from pathlib import Path
+sys.path.insert(0, str(Path(__file__).parent / "src"))
+
+from monarchmoney import MonarchMoney
+from monarch_mcp_server.secure_session import secure_session
+
+async def main():
+    if len(sys.argv) < 2:
+        print("Usage: uv run python save_token.py YOUR_TOKEN_HERE")
+        sys.exit(1)
+    token = sys.argv[1].strip()
+    mm = MonarchMoney(token=token)
+    print("Testing connection...")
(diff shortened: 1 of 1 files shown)


robcerda:
Closing for two reasons.

Security regression. Accepting the token as sys.argv[1] makes it appear in shell history (.bash_history, .zsh_history) and briefly visible in ps output to other local users. getpass and stdin reads do not have either property. The terminal paste-into-getpass UX problem is real, but the fix should pipe the token via stdin (cat token.txt | uv run python save_token.py, or read -s; echo "$TOK" | ...) rather than argv.

Auth path is currently broken upstream. MonarchMoney(token=...) sends Authorization: Token, which Monarch's API no longer accepts as of May 2026 (see #71 for the analysis and #59 for the cookie-based replacement). Even with the security fix, the saved token would not authenticate against the live API today.

Happy to look at a stdin-based variant once the cookie auth path lands.

PR #46, opened 2026-04-23
Rebase yonib05/all-features onto main (refresh of #29)

madisonrickert:
Refresh of the stalled PR #29 from @yonib05, rebased onto current `main` and reconciled with the upstream changes that have landed since.

All 21 of @yonib05's original commits are preserved with their original authorship. Conflict resolution and two cleanup commits are authored by me.

## What this picks up since #29 was opened

These five upstream commits have landed in `main` and are reconciled into the modular `tools/` layout:

- `bf3e604` (PR #35, mine) — `create_transaction`/`update_transaction` SDK signature fix; carried forward into `tools/transactions.py`
- `037876d` (PR #41, @kylelol) — `get_transaction_categories`, `get_transaction_tags`, `set_transaction_tags`, `create_transaction_tag`, `categorize_transaction`
- `a8f30e9` (PR #41, @kylelol) — `get_transaction_category_groups`, `create_transaction_category`
- `ade9edc` (PR #41, @kylelol) — `add_transaction_tag` (union-preserving)
- `90025dd` (PR #42, @rich4756) — enhanced `get_transactions`: new response fields (`needs_review`, `notes`, `category_id`, `account_id`, `tags`, `review_status`, `is_recurring`, `is_split_transaction`, `hide_from_reports`), new filter params (`search`, `category_ids`, `account_ids`, `tag_ids`, `has_notes`, `is_split`, `is_recurring`), and the `account_id` → `account_ids` list-wrapping bugfix

## Naming reconciliation

@yonib05's earlier branch added overlapping tools under shorter names. Since @kylelol's names were merged into `main`, the canonical names from `main` are the only ones that remain — @yonib05's shorter versions are dropped:

| Dropped (was on yonib05's branch) | Kept (already on `main`) |
|---|---|
| `get_categories` | `get_transaction_categories` |
| `get_category_groups` | `get_transaction_category_groups` |
| `get_tags` | `get_transaction_tags` |
| `create_tag` | `create_transaction_tag` |
| `set_transaction_category` | `categorize_transaction` |

`set_transaction_tags` was unchanged — the name was the same in both branches.

## Two regressions fixed

Surfaced while porting tests:

1. `tools/accounts.py:get_accounts` was missing the `is_hidden` field that the pre-refactor `server.py` returned.
2. `tools/budgets.py:get_budgets` had reimplemented the buggy formatter that was deliberately replaced with a raw-response in #39 (wrong response key produced empty results). Swapped for the raw-response pattern.

## Tests

[...]

PR #71, opened 2026-06-23
Support current Monarch auth flow

thewoolleyman:
## Summary

Monarch's auth flow has changed in ways that break this MCP server's current setup:

- API calls should use `https://api.monarch.com`, not the old `https://api.monarchmoney.com` host.
- New/unrecognized sessions may require email OTP even when account MFA is disabled.
- The existing `monarchmoney` package misclassifies the email-code response as a generic 403/MFA-style failure.
- Reloading a saved token needs the same device UUID metadata used during login; saving only the raw token can produce `401 Unauthorized` on GraphQL calls.
- The upstream budget query asks for category-group fields that Monarch no longer accepts for this account, so `get_budgets` can fail even after auth is fixed.

This PR adds a small compatibility layer around the existing `monarchmoney` dependency instead of rewriting the MCP tools:

- patches `MonarchMoneyEndpoints.BASE_URL` to the current Monarch API host
- adds current web login payload fields: `supports_email_otp`, `supports_recaptcha`, and `email_otp`
- detects email OTP responses separately from TOTP MFA responses
- updates `login_setup.py` to prompt for email verification codes and save the resulting session to keyring
- stores token plus device UUID in keyring, while still accepting legacy raw-token entries
- creates stored-token clients with the current API host and saved device metadata
- keeps credential login inside `login_setup.py`; MCP tool calls now use only an existing keyring session and do not trigger noninteractive password-login attempts
- uses a narrower budget GraphQL query that avoids stale category-group fields and returns current-month category budget rows
- suppresses verbose GraphQL transport logging to avoid logging account payloads
- updates README auth/session docs

## Upstream context

Related upstream dependency issues:

- `hammem/monarchmoney#184`: API domain changed to `api.monarch.com`
- `hammem/monarchmoney#196`: first non-interactive auth attempt returns 429; commenters report the new host fixes that part
- `hammem/monarchmoney#137`: app update warning / auth 403 is misclassified as MFA

The TypeScript `@hakimelek/monarchmoney` package documents the same current behavior: Monarch may require email OTP for new devices/sessions even when MFA is off, and the OTP is submitted as `email_otp`.

## Verification

```bash
uv run --with-editable . --with pytest python -m pytest -q
uv run python -m py_compile src/monarch_mcp_server/config.py src/monarch_mcp_server/monarch_auth.py src/monarch_mcp_server/secure_session.py src/monarch_mcp_server/server.py login_setup.py
uv run --with 'mcp[cli]' --with-editable . mcp run src/monarch_mcp_server/server.py
```

Additional local smoke verification completed:

- `login_setup.py` loaded credentials from `.env.local`
- email OTP flow completed successfully
- setup fetched accounts during verification
- session was saved to keyring
- a fresh Python process loaded the saved keyring session successfully
- fresh-process smoke checks passed for auth status, accounts, one transaction, budgets, cashflow, and holdings
- Codex MCP tool calls for auth status, accounts, transactions, and cashflow succeeded after keyring renewal

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,8 +153,10 @@ uv run python login_setup.py
 
 Follow the prompts:
 - Enter your Monarch Money email and password
+- Provide the email verification code if Monarch sends one (this can happen for a
+  new device or session even when MFA is not enabled)
 - Provide 2FA code if you have MFA enabled
-- Session will be saved automatically
+- The session token is saved automatically in your system keyring
 
 ### 3. Start Using
 
@@ -231,6 +233,7 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 
 ### 🔐 Secure Authentication
 - **One-Time Setup**: Authenticate once, use for weeks/months
+- **Email OTP Support**: Handles Monarch's email verification flow for new devices/sessions
 - **MFA Support**: Full support for two-factor authentication
 - **SSO/Google sign-in**: Use `monarch_login_with_token` to paste a session token from your browser
 - **Session Persistence**: No need to re-authenticate frequently
@@ -402,10 +405,16 @@ If you see "Authentication needed" errors:
 2. Restart Claude Desktop or Claude Code
 3. Try using a tool like `get_accounts`
 
+### Email Verification Required
+Monarch may require an email one-time code for a new device or session, even if MFA is not enabled. If you see an email-code prompt:
+1. Check the email address on your Monarch account
+2. Enter the one-time code in `login_setup.py`
+3. Let the script finish so it can save the reusable token to your system keyring
+
 ### Session Expired
 Sessions last for weeks, but if expired:
 1. Run the same setup command again: `python login_setup.py` (or `uv run python login_setup.py`)
-2. Enter your credentials and 2FA code
+2. Enter your credentials and any requested email verification or 2FA code
 3. Session will be refreshed automatically
 
 ### `'Context' object has no attribute 'elicit'`
@@ -419,6 +428,7 @@ Then fully quit and reopen Claude Desktop or Claude Code so it relaunches the se
 
 ### Common Error Messages
 - **"No valid session found"**: Run `python login_setup.py` (or `uv run python login_setup.py`) 
+- **"Monarch sent a one-time code to your email"**: Run `python login_setup.py` and complete email verification
 - **"Invalid account ID"**: Use `get_accounts` to see valid account IDs
 - **"Date format error"**: Use YYYY-MM-DD format for dates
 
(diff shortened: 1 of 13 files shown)


robcerda:
Thanks for the thorough write-up. Holding this open rather than merging because there are blockers that need a coordinated decision.

Security concerns.

1. load_dotenv is reintroduced and reads .env, .env.local, and CWD .env. PR #61 (read-only hardening) is removing dotenv loading specifically because a stray .env next to the source tree is an authentication-bypass surface. Want to settle that before reintroducing it.

2. README promotes MONARCH_PASSWORD in .env as a setup path. Plaintext password on disk is exactly what #61 is hardening against.

3. monarch-client-version v1.0.1668 is hardcoded. It will rot silently against a server-side check. At minimum extract it to a named constant with a comment about when it was captured and where to recheck.

4. Log suppression at logging.getLogger("gql.transport.aiohttp").setLevel(logging.WARNING) is reasonable to avoid logging account payloads but should be documented inline so future debugging does not silently drop transport errors.

Architectural conflict.

5. This edits the monolithic server.py, but server.py was refactored into the tools/ package in PR #46 back in April. The PR is CONFLICTING and the conflict is not trivial. It needs a rebase onto the modular layout.

6. Three competing auth approaches now exist for the same May-2026 Monarch change: #59 (cookie auth via session_id + csrftoken + x-csrftoken header), #61 (read-only by default + auth tools hard-disabled), and this PR (api.monarch.com base URL + email OTP + token + device-uuid). They diverge on session model. Need a maintainer call on which line to pursue.

Breaking change to existing callers.

7. get_budgets response shape changes from name/amount/spent/remaining/category/period to id/name/planned/actual/remaining/category_group/month. Any client (or LLM prompt) referring to the old field names breaks.

The email-OTP detection, the api.monarch.com host fix, and the device-uuid plus token storage are all good ideas. The form that lands here will most likely fold those into the cookie-auth path from #59 once that drops draft.

robcerda:
This is conflicting against current main because of the April refactor of server.py into the tools/ package. Could you rebase onto current main so we can evaluate it cleanly?

thewoolleyman:
Thanks for the detailed review — that all makes sense. I've force-pushed a version rebased onto the modular `tools/` layout (so it's no longer the monolithic `server.py` conflict), and addressed the mechanical blockers:

**Rebase (#5).** Ported onto the `tools/` package. `server.py` stays the re-export shim. The auth compat lives in a new `monarch_auth.py`; budget changes are in `tools/budgets.py`; the device-uuid storage folds into the post-#72 keyring/file-fallback `secure_session.py`.

**dotenv (#1).** Dropped entirely — no `config.py`, and I also removed the existing `load_dotenv()` from `app.py`. There's no `.env`/`.env.local`/CWD loading anywhere in this branch now. (This overlaps with what #61 is doing; happy to defer that one line to #61 if you'd rather keep the surface area of this PR tighter.)

**README password (#2).** Removed — no `MONARCH_PASSWORD` setup path. README now documents keyring-only storage plus the email-OTP flow.

**Client version (#3).** Extracted to a `MONARCH_CLIENT_VERSION` constant with a comment noting when it was captured (2026-06-23) and how to recheck it from the web app's request headers.

**Log suppression (#4).** Moved to `app.py` with an inline comment explaining it's to keep account payloads out of logs, that transport errors still surface, and how to re-enable for debugging.

I also removed the env-credential auto-login from `client.py` so MCP tool calls only ever use an existing keyring session — they never initiate a noninteractive password login. That lines up with the hardening direction in #61.

**Budget shape (#7).** Worth a second look: on current `main`, `get_budgets` already returns the raw SDK `get_budgets()` payload (not the `name/amount/spent/remaining/...` shape from my original monolithic diff), so there isn't a stable documented shape being broken anymore. The new query returns one row per category/month with `id/name/planned/actual/remaining/category_group/month`, documented in the tool docstring, and keeps the `start_date`/`end_date` params. If you'd prefer, I'm happy to split the budget fix into its own PR so the auth fix can land independently.

**Auth strategy (#6).** This is the real open question for you: you mentioned folding these into #59's cookie-auth path once it dropped draft, but #59 is now closed. So there's no longer a converging target. The pieces here — `api.monarch.com` host, email-OTP detection, and token+device-uuid storage — are deliberately modular, so I can graft them onto whichever session model you want to standardize on. Which line do you want to pursue now that #59 is out?

PR #78, opened 2026-07-09
fix: correct Monarch domain to monarch.com + graceful fallback for clients without elicitation

shmuelsash:
## What

Two correctness/robustness fixes:

1. **Domain migration in user-facing text.** Monarch retired `app.monarchmoney.com` in favour of `app.monarch.com`. The token instructions (`TokenForm` description and the `monarch_login_with_token` docstring) still point users at the old domain, where the `token` local-storage key no longer lives. Updated to `app.monarch.com`.

2. **Graceful handling for MCP clients without elicitation.** Some clients don't implement the elicitation protocol and return `-32601 Method not found` when `ctx.elicit(...)` is called. Today that surfaces as an opaque error. This catches it in the login, MFA, and token-login flows and returns a clear message telling the user to use `monarch_login_with_token` instead (with step-by-step token retrieval).

## Notes

- No behavioral change for clients that *do* support elicitation — the try/except only intercepts the `-32601` case and re-raises anything else.
- No new dependencies.

### DIFF
--- a/src/monarch_mcp_server/auth.py
+++ b/src/monarch_mcp_server/auth.py
@@ -22,6 +22,14 @@ _UPGRADE_HINT = (
     "`python login_setup.py` from the repo to authenticate via terminal."
 )
 
+_CLIENT_NO_ELICITATION = (
+    "Your MCP client does not support the elicitation protocol "
+    "(it returned -32601 Method not found). "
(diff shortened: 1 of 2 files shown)


robcerda:
Both changes are correct: the docstrings still pointed at the old monarchmoney.com host, and a client without elicitation support currently surfaces a raw JSON RPC error instead of a usable next step. Merging.

PR #79, opened 2026-07-10
Add remote HTTP serve mode: bearer-token auth, headless login, and Claude connector support

arditti:
## Summary

- New `serve` subcommand/run mode exposing the server over MCP streamable
  HTTP (in addition to the existing stdio mode — unaffected), authenticated
  with a static bearer token, so it can be deployed on a VPS/Raspberry
  Pi/Docker host and reached remotely behind a reverse proxy or tunnel.
- Docker packaging (`Dockerfile`, `docker-compose.yml`, `.dockerignore`)
  and a Cloudflare Tunnel example for running the server remotely.
- `mcp` upgraded to `>=1.28,<2` so streamable HTTP sessions are bound to
  their credentials (fixes a session-replay gap present in 1.13.1).
- Bearer token verification uses constant-time digest comparison and a
  global failure throttle to resist timing/brute-force attacks; the
  interactive login tools are only registered in stdio mode — credentials
  never flow through an LLM in serve mode.
- `login-cookies` / `login-password` non-interactive subcommands: seed a
  Monarch session from a browser Cookie header or email+password (with
  email-OTP/TOTP support) piped over stdin, for headless hosts where
  `login_setup.py`'s interactive prompts aren't available (e.g. `docker
  exec` from a CI runner). A supplied one-time code now rides the *first*
  login request — a bug where a no-code probe attempt would trigger a
  fresh OTP email and invalidate the code just supplied is fixed.
- `/` (root): an optional self-service copy-paste client-config page —
  Claude Code command, Claude Desktop config JSON, a ready-to-paste
  connector URL, and the raw bearer token. It has no auth of its own, so
  deployers should put it behind their own access control (reverse-proxy
  auth, an identity-aware proxy, etc.) if serving to more than themselves.
- `?token=` query-param bearer-token fallback: Claude's native "Add custom
  connector" UI (web/desktop/mobile) only accepts a URL, not a custom
  header, and leaving its OAuth fields blank triggers a full OAuth
  discovery/dynamic-registration flow this static-token resource server
  doesn't implement. Implemented as a small ASGI middleware wrapping
  FastMCP's app (checked only when no Authorization header is present, so
  an explicit header always wins) — the `mcp` SDK itself is untouched.

## Design docs included

- `docs/superpowers/specs/2026-07-08-remote-http-phase1a-design.md` —
  design spec (reviewed for security, SDK feasibility, and architecture)
- `docs/superpowers/plans/2026-07-08-remote-http-phase1a.md` —
  implementation plan

## Test plan

- [x] 281 tests passing: run-mode resolution, serve config, bearer token
  verifier, HTTP auth-boundary (real subprocess server, real HTTP),
  secure session, login-cookies/login-password (including the OTP-timing
  fix), the setup page, and the query-token middleware (unit + live-server
  integration).
- [x] `scripts/e2e_http.py` — end-to-end check against a running server
  using the real MCP client.
- [x] Manually verified against a real deployment: Docker Compose behind a
  reverse proxy/tunnel, live Monarch login, and the native-connector
  `?token=` flow all confirmed working end to end.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,3 +7,6 @@ docs
 scripts
 .pytest_cache
 .mypy_cache
+.github
+.gitignore
+.superpowers
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+.git
+.venv
+__pycache__
+*.pyc
+tests
+docs
+scripts
+.pytest_cache
+.mypy_cache
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN pip install --no-cache-dir .
 
 # Non-root user; HOME=/data puts the file-fallback session store
 # (~/.monarch-mcp-server/token) on the persistent volume.
-RUN useradd --create-home --uid 1000 monarch \
+RUN useradd --no-create-home --uid 1000 monarch \
     && mkdir -p /data \
     && chown monarch:monarch /data
 ENV HOME=/data \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM python:3.12-slim
+
+WORKDIR /app
+
+COPY pyproject.toml README.md LICENSE ./
+COPY src ./src
+COPY login_setup.py ./
+
+RUN pip install --no-cache-dir .
+
+# Non-root user; HOME=/data puts the file-fallback session store
+# (~/.monarch-mcp-server/token) on the persistent volume.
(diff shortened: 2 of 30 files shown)


robcerda:
This is careful work, but exposing a network listener with bearer auth and a headless login page from a process that holds live Monarch credentials is a large new attack surface, and at 3400 lines across 30 files it is not something I can review to the standard that surface needs. Keeping the server stdio only for now, so I am closing this rather than leaving it open indefinitely.

PR #81, opened 2026-07-16
Encrypt file-fallback session at rest with Windows DPAPI

### LINKED ISSUE #80: Windows: keyring probe passes but real cookie-session save silently falls back to a plaintext file
## Summary

On Windows, `_keyring_available()` reports the keyring as usable, the server logs `🔐 Using system keyring for token storage`, and then the **actual** session save silently falls back to a **plaintext** file — because a real cookie-mode session is too large for Windows Credential Manager. Users who believe their credentials live in the encrypted OS vault instead have their live Monarch session cookies sitting in a plaintext file, with no clear signal that this happened.

## Root cause

`_keyring_available()` (in `src/monarch_mcp_server/secure_session.py`) probes the backend by round-tripping a **1-byte** sentinel (`"1"`):

```python
keyring.set_password(KEYRING_SERVICE, _PROBE_USERNAME, "1")
```

That always succeeds on Windows. But Windows Credential Manager (`keyring.backends.Windows.WinVaultKeyring`) has a hard cap on the credential blob size. A cookie-mode session blob — `{"auth_mode": "cookie", "device_uuid": ..., "cookies": {session_id, csrftoken, cf_clearance, __cf_bm, ...}}` — is ~950 chars, which exceeds the cap. So `save_session_blob()` throws inside the `try`, hits the `except`, and quietly writes the plaintext file instead:

```python
except Exception as e:
    logger.warning(f"⚠️  Keyring save failed, falling back to file: {e}")
self._save_token_file(blob)   # plaintext
```

The only trace is a single `WARNING` line that's easy to miss; `check_auth_status` then also reports "no token found in keyring" even though auth works via the file, which compounds the confusion.

## Reproduction / evidence

On Windows 11 (Credential Manager backend), `CredWrite` starts failing well below the documented limit:

```
  600 chars: OK
  700 chars: FAIL -> (8, 'CredWrite', 'Not enough memory resources are available to process this command.')
 1500 chars: FAIL -> (1783, 'CredWrite', 'The stub received bad data.')
```

The documented Windows limit is 2560 UTF-16 chars; some machines are stricter. Either way, a cookie session blob exceeds it. This is the same platform limitation OpenAI Codex hit for OAuth tokens on Windows: openai/codex#10353 (closed, no platform-level fix — the token simply cannot go in the vault).

## Impact

- **Silent security downgrade on Windows:** live session cookies stored as plaintext at rest, while the logs claim the keyring is in use.
- Affects the recommended auth path (browser session cookies), i.e. most Windows users.

## Suggested fixes

1. **Make the availability probe size-aware** so the "using keyring" decision reflects what will actually be stored (probe with a representative-size value, or catch the size failure at decision time). At minimum, make the fallback log unmistakable.
2. **Encrypt the file fallback at rest.** On Windows, DPAPI (`win32crypt.CryptProtectData`, already available via the `pywin32` dependency) gives the same per-user protection Credential Manager uses, with no size cap. No-op on other platforms.

I'll open a PR implementing (2), with transparent migration of existing plaintext files.

chirstius:
Closes #80.

## Problem

On Windows, a cookie-mode session blob (~950 chars) exceeds Windows Credential Manager's credential-size cap, so `save_session_blob()` throws, hits the `except`, and silently writes the session to a **plaintext** file — even though `_keyring_available()` (which probes with a 1-byte value) reported the keyring as usable and the logs say the keyring is in use. The result is live Monarch session cookies stored in cleartext at rest, without the user realizing it. See #80 for the full analysis and repro.

## Change

Encrypt the file fallback at rest on Windows using DPAPI (`win32crypt.CryptProtectData`), scoped to the current user. DPAPI is the same per-user protection Credential Manager uses under the hood, but without the blob-size cap, so it comfortably holds a cookie session.

- New helpers `_dpapi_available()` / `_dpapi_encrypt()` / `_dpapi_decrypt()`; encrypted values are stored as `DPAPI:<base64>`.
- `_save_token_file()` encrypts when DPAPI is available.
- `_load_token_file()` decrypts `DPAPI:`-prefixed content and **transparently migrates** any existing legacy plaintext file to encrypted on first read.
- **No-op on non-Windows platforms** — the file stays exactly as before, so Linux/macOS/WSL behavior is unchanged.
- No new dependencies: `pywin32` (which provides `win32crypt`) already ships transitively with the existing deps on Windows.

This intentionally does **not** touch the keyring path or the `_keyring_available()` probe — it only hardens the fallback that's already being used. Making the probe size-aware (so the "using keyring" log is accurate) is worth a follow-up but is orthogonal to encrypting data at rest.

## Testing

On Windows 11 (Python 3.13, Credential Manager backend):

- Existing plaintext `~/.monarch-mcp-server/token` auto-migrated to `DPAPI:`-prefixed ciphertext on load.
- Round trip verified: encrypt → read back → decrypt → live `get_accounts()` call succeeds (41 accounts).
- On-disk file confirmed to be ciphertext (no cookie values in cleartext).

## Notes

Same class of platform limitation OpenAI Codex hit for Windows OAuth tokens (openai/codex#10353) — there's no way to make a large credential fit the vault, so encrypting the fallback is the practical fix.

### DIFF
--- a/src/monarch_mcp_server/secure_session.py
+++ b/src/monarch_mcp_server/secure_session.py
@@ -5,10 +5,12 @@ Uses the system keyring when available, with an automatic file-based
 fallback for environments without a keyring backend (e.g. WSL, headless Linux).
 """
 
+import base64
 import json
 import logging
 import os
 import stat
+import sys
 from pathlib import Path
 from typing import Any, Dict, Optional
 from monarchmoney import MonarchMoney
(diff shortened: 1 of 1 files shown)


robcerda:
Rebased your commit onto main and merged it in #123, since #113 had changed the same function and this needed to sit alongside the atomic write and 0600 creation fixes. Authorship preserved, thanks for the DPAPI work.

PR #83, opened 2026-07-28
Fix: login_setup.py deletes existing session before new login is verified

peter216:
*Beep boop, I am Claude Code 🤖, my user has reviewed and approved the following written by me:*

**Summary**

`login_setup.py`'s `main()` calls `secure_session.delete_token()` unconditionally as the very first step, before the new login attempt (cookies, password+MFA, or legacy token) is tried, tested against `get_accounts()`, or saved:

```python
try:
    secure_session.delete_token()
    print("🗑️ Cleared existing secure sessions")

    print("\nHow do you sign in to Monarch Money?")
    ...
```

If anything fails after that point — a bad cookie paste, a Cloudflare captcha, a 401 on the connection test, an exception during save — the script prints an error and exits, but the previously-working session has already been wiped. The user ends up strictly worse off than before running the script: no working session at all, where they had one before.

**Why the delete isn't needed**

`save_authenticated_session()` → `save_session_blob()` already overwrites the stored keyring entry (`keyring.set_password` on the same service/username) or the file-fallback (`Path.write_text`, which truncates), and calls `_cleanup_old_session_files()` itself on the success path. The upfront `delete_token()` call has no effect on a successful run — its only observable effect is destructive, and only on the failure path.

**Fix**

Removed the premature `delete_token()` call. The old session is now only ever replaced once a new one is confirmed working (passes the `get_accounts()` test) and saved successfully — the existing overwrite-on-save behavior handles the replacement safely.

**Testing**

Read through the full call graph (`login_setup.py` → `secure_session.save_authenticated_session` → `save_session_blob`) to confirm the overwrite behavior on the success path is unaffected. Didn't have Monarch credentials to exercise the live login flow end-to-end, so I'd appreciate a maintainer sanity-check on a real account if that's a concern — the diff itself is a 3-line deletion with no other logic changes.

*Beep boop, Claude Code 🤖 out!*

### DIFF
--- a/login_setup.py
+++ b/login_setup.py
@@ -125,9 +125,6 @@ async def main():
         print(f"⚠️  Could not check version: {e}")
 
     try:
-        secure_session.delete_token()
-        print("🗑️ Cleared existing secure sessions")
-
         print("\nHow do you sign in to Monarch Money?")
         print(
             "  1) Session cookies from browser   "
--- a/src/monarch_mcp_server/tools/auth.py
+++ b/src/monarch_mcp_server/tools/auth.py
@@ -66,9 +66,18 @@ async def monarch_logout() -> str:
 async def check_auth_status() -> str:
     """Check if already authenticated with Monarch Money."""
     try:
-        token = secure_session.load_token()
-        if token:
-            status = "✅ Authentication token found in secure keyring storage\n"
+        # load_token() only ever inspects the "token" key, so it reports a
+        # false negative for cookie-mode sessions (auth_mode="cookie"), which
+        # carry a "cookies" dict and no top-level token. Check the full
+        # session so a valid cookie-only login — the recommended, long-lived
+        # auth path — is correctly recognized as authenticated.
+        session = secure_session.load_session()
+        has_token = bool(session and isinstance(session.get("token"), str) and session.get("token"))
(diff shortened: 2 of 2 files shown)


peter216:
*Beep boop, I am Claude Code 🤖, my user has reviewed and approved the following written by me:*

Folded in a second, closely-related fix (commit `109ee80`): after re-logging in via the cookie method, `check_auth_status` still reported "No authentication token found" even though the new session was fully valid — confirmed live, `get_accounts()` returned real data through the exact same session `check_auth_status` said was missing.

Root cause: `check_auth_status` calls `secure_session.load_token()`, which only ever inspects the top-level `"token"` key. A cookie-mode session (`auth_mode: "cookie"`) stores its credentials under a `cookies` dict instead, with no top-level token — a false negative on precisely the recommended, long-lived login path this PR is already about. Now checks the full session via `load_session()` and treats either a non-empty `token` or a non-empty `cookies` dict as authenticated.

Same theme as the first commit (session state getting mishandled around the cookie-auth path), so keeping it as one PR rather than opening a second.

*Beep boop, Claude Code 🤖 out!*

robcerda:
Correct catch, the upfront delete only ever mattered on the failure path. Landed the login_setup.py half in #124 with attribution, since the auth.py half conflicted with #110 which fixed the same cookie mode status bug.

PR #86, opened 2026-08-07
Fix delete_transaction_rule reporting successful deletions as failures

ahnj:
## Problem

`delete_transaction_rule` returns `{"success": false, "message": "Rule was not deleted"}` for deletions that actually succeed.

Monarch's `deleteTransactionRule` mutation returns `deleted: false` with no errors even when the rule *was* removed. The current guard treats that flag as authoritative:

```python
if delete_result.get("deleted") is False:
    return json_success({"success": False, "message": "Rule was not deleted"})
```

This is more than cosmetic — the tool reports a completed destructive operation as not having happened, which invites the caller to retry or to act on state they believe is still present.

## Verification against the live API

Deleted four real rules. Every call returned `deleted: false`; all four were genuinely gone, confirmed by re-querying the rule list (the IDs were absent) and by the total count dropping from 322 to 318.

Genuine failures never reach that branch. Deleting a non-existent rule raises `TransportQueryError` from the GraphQL layer:

```
gql.transport.exceptions.TransportQueryError:
{'message': 'Not found', 'path': ['deleteTransactionRule']}
```

which the surrounding `except Exception` already turns into an error result.

## Change

Drop the `deleted` check. The explicit `errors` payload check is left in place as the only trustworthy in-payload failure signal.

## Tests

`test_delete_rule_not_found` passed before and after because it mocks `deleted: false` *together with* an errors object, so the errors branch catches it first — the real-world shape was never covered.

Added `test_delete_rule_success_with_false_deleted_flag` for `deleted: false` + `errors: None`. It fails before this change and passes after. Full suite: 223 passed.

https://claude.ai/code/session_01KTJGjJh4qNHBGWdZW3UDPo

### DIFF
--- a/src/monarch_mcp_server/tools/rules.py
+++ b/src/monarch_mcp_server/tools/rules.py
@@ -421,19 +421,21 @@ async def delete_transaction_rule(rule_id: str) -> str:
             variables={"id": rule_id},
         )
 
-        # Monarch's deleteTransactionRule can return a payload where the
-        # `deleted` flag is absent/null even when the deletion succeeded, which
-        # previously produced a false "Unknown error". Treat an explicit errors
-        # payload (or deleted == False) as failure; otherwise the mutation was
-        # accepted and the rule is gone.
+        # Monarch's deleteTransactionRule returns an unreliable `deleted` flag:
+        # verified against the live API, it comes back `false` even when the rule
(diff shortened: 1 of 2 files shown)


robcerda:
The live API evidence is convincing, and reporting a completed destructive operation as failed is worse than the false negative it replaced. Merging.

PR #87, opened 2026-08-10
feat: migrate to mcp Python SDK >=2.0

dcode:
## Summary
- The `mcp` Python SDK's 2.0 release renamed `mcp.server.fastmcp.FastMCP` to `MCPServer` and relocated it (along with `Context`) to `mcp.server.mcpserver`. The old `fastmcp` module was removed outright, not aliased, so this is a hard cutover rather than an additive change.
- Updates all imports (`app.py`, `auth.py`, `tools/auth.py`) and lingering "FastMCP" references in comments/docstrings/README to the new module and class name.
- Bumps the `mcp[cli]` dependency floor to `>=2.0.0` in `pyproject.toml` and `requirements.txt`, and regenerates `uv.lock` (`uv lock --upgrade-package mcp`), which pulled in a few transitive bumps (pydantic, httpx→httpx2, added `mcp-types`, `opentelemetry-api`, `pyjwt`, `truststore`).

## Test plan
- [x] `uv sync --extra dev` installs cleanly against mcp 2.0.0
- [x] `python -c "from monarch_mcp_server.app import mcp; print(type(mcp))"` → `mcp.server.mcpserver.server.MCPServer`
- [x] `uv run pytest -q` → 221/222 passing (the one failure, `test_no_session_returns_none`, is pre-existing on `main` and unrelated to this change — verified by running it against the original code/lockfile)

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,7 +451,7 @@ Then fully quit and reopen Claude Desktop or Claude Code so it relaunches the se
 monarch-mcp-server/
 ├── src/monarch_mcp_server/
 │   ├── __init__.py
-│   ├── app.py             # FastMCP app instance and entry point
+│   ├── app.py             # MCPServer app instance and entry point
 │   ├── client.py          # Cached MonarchMoney client factory
 │   ├── monarch_auth.py    # Current Monarch auth compatibility (host, email OTP, device-uuid)
 │   ├── secure_session.py  # Keyring-backed token storage (file fallback)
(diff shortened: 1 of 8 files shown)


robcerda:
Appreciate the thorough migration, but dropping mcp 1.x support would break every existing pinned install. #108 landed a compat shim that works on both majors, verified with the full suite under 1.13.1 and 2.1.1.

PR #88, opened 2026-08-11
Read browser cookie from a config file (fixes #82)

### LINKED ISSUE #82: macOS: cookie login hangs forever — getpass exceeds 1024-byte MAX_CANON tty limit
## Summary

On macOS, option 1 (browser session cookies) in `login_setup.py` can never succeed. The script hangs forever at the `Paste the Cookie header value:` prompt, and Enter does nothing.

The cause is a hard tty limit, not a bad cookie or a slow network call.

## Cause

`login_setup.py:42` reads the cookie with `getpass.getpass()`:

```python
cookie_string = getpass.getpass("Paste the Cookie header value: ").strip()
```

`getpass` reads a line from `/dev/tty` in **canonical mode**. Canonical-mode input is capped at `MAX_CANON` bytes per line, and on Darwin that is **1024**. A real Monarch `Cookie` header is several KB (Segment `ajs_*`, Osano consent, Cloudflare, plus the auth cookies), so the line never terminates: the terminal emits BEL for every byte past the limit and discards it. `readline()` blocks indefinitely.

Linux sets `MAX_CANON` to 4096, which is probably why this hasn't shown up in testing.

## Reproduction

This reproduces the line-discipline behavior directly, no Monarch account needed (macOS 26.5.2, CPython 3.13.14):

```python
import os, pty, termios, time

def trial(n):
    pid, fd = pty.fork()
    if pid == 0:
        f = open('/dev/tty', 'r')
        attrs = termios.tcgetattr(f)
        attrs[3] = attrs[3] & ~termios.ECHO   # what getpass does; ICANON stays on
        termios.tcsetattr(f, termios.TCSAFLUSH, attrs)
        line = f.readline()
        os.write(1, b"GOT:%d\n" % len(line))
        os._exit(0)
    time.sleep(0.4)
    os.write(fd, b"C" * n + b"\n")
    time.sleep(0.6)
    os.set_blocking(fd, False)
    try:
        out = os.read(fd, 65536)
    except Exception as e:
        out = repr(e).encode()
    os.kill(pid, 9); os.waitpid(pid, 0)
    print(f"sent {n:5d} bytes -> {out[:40]!r}")

for n in (900, 1023, 1030, 4000):
    trial(n)
```

Output:

```
sent   900 bytes -> b'GOT:901\r\n'
sent  1023 bytes -> b'GOT:1024\r\n'
sent  1030 bytes -> b'\x07\x07\x07\x07\x07\x07\x07'
sent  4000 bytes -> b'\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07\x07...'
```

1023 bytes works, 1030 does not. Past the limit the read never completes, which matches the hang. Interrupting produces:

```
File "login_setup.py", line 42, in _login_with_cookies
    cookie_string = getpass.getpass("Paste the Cookie header value: ").strip()
...
File ".../getpass.py", line 146, in _raw_input
    line = input.readline()
KeyboardInterrupt
```

This is also user-hostile to diagnose, because `getpass` echoes nothing — there is no feedback distinguishing "waiting for input" from "input silently dropped."

## Suggested fixes

Any of these avoids the canonical-mode limit:

1. **Read the clipboard.** The user has just copied the value anyway. `subprocess.run(["pbpaste"], ...)` on macOS, `xclip -o`/`wl-paste` on Linux, `Get-Clipboard` on Windows. Best UX by far: no prompt, no paste into a tty at all.
2. **Read from a file.** Prompt for a path, or accept `--cookie-file`. Also keeps the secret out of shell history.
3. **Read from an env var**, e.g. `MONARCH_COOKIE`.
4. **Turn off `ICANON` for the read** and accumulate until newline, instead of relying on the line discipline.

Option 4 keeps the existing UX but is the fiddliest; option 1 or 2 is probably the smallest safe change. A fallback chain (try the tty read, and if it returns empty or the process is on Darwin, fall back to clipboard/file) would preserve current behavior on Linux.

It would also help to validate the parsed result and tell the user what was actually captured — printing cookie *names* and byte count (never values) makes "I grabbed the wrong request in DevTools" immediately obvious, since the analytics-only cookies from a non-`api.monarch.com` request are a separate easy mistake.

I worked around it locally with a small clipboard-based script and authenticated successfully on the first try, so the cookie auth path itself works fine — it is only the input method that is broken. Happy to send a PR if you have a preference among the above.

## Environment

- macOS 26.5.2 (Darwin 25.5.0), arm64
- Python 3.13.14
- monarchmoneycommunity 1.4.0
- monarch-mcp-server @ ca6c159 (main, cloned 2026-07-26)

MikeDacre:
Interactive pastes via getpass()/input() are silently truncated at the POSIX terminal's canonical-mode input buffer (MAX_CANON, 1024 bytes on macOS/Linux). Real Monarch cookie headers routinely exceed 1 KB, so the paste-at-prompt flow fails with a confusing auth error.

login_setup.py now reads the cookie from a platform-appropriate config file when present, falling back to the interactive prompt with a warning about the truncation limit:

  - macOS/Linux: ~/.config/monarch-mcp/cookie.txt (respects XDG_CONFIG_HOME)
  - Windows: %APPDATA%\monarch-mcp\cookie.txt
  - Any platform: override with MONARCH_MCP_COOKIE_FILE

README updated with per-platform setup instructions. Verified against the live API (get_accounts) on macOS; Windows path resolution covered by mocked-platform checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,9 +152,35 @@ Long-lived sessions, supports SSO accounts, and sidesteps Cloudflare CAPTCHA gat
 2. Open DevTools (F12) → Network tab.
 3. Click any request whose Name starts with `graphql` (or any request to `api.monarch.com`).
 4. Scroll to Request Headers, find the `cookie:` header, and copy the full value.
-5. Paste it into the prompt.
+5. Save it to the cookie file for your platform (recommended), then re-run the script — it reads the file automatically:
 
-The script verifies the cookies against the live API before saving them to your system keyring.
+   **macOS / Linux** — `~/.config/monarch-mcp/cookie.txt` (respects `$XDG_CONFIG_HOME`):
+
+   ```bash
+   mkdir -p ~/.config/monarch-mcp
+   # paste the cookie value into the file with your editor, then:
+   chmod 600 ~/.config/monarch-mcp/cookie.txt
+   ```
+
+   **Windows** — `%APPDATA%\monarch-mcp\cookie.txt`:
+
+   ```powershell
+   New-Item -ItemType Directory -Force "$env:APPDATA\monarch-mcp" | Out-Null
+   notepad "$env:APPDATA\monarch-mcp\cookie.txt"   # paste the cookie value, save, close
+   ```
+
+   Files under your user profile are already ACL-restricted to your account on Windows; no `chmod` equivalent is needed for typical single-user machines.
+
+   To use a different location on any platform, set the `MONARCH_MCP_COOKIE_FILE` environment variable to the full path.
+
+   Alternatively, paste the value at the interactive prompt — but note that
+   POSIX terminals silently truncate pasted input at the canonical-mode
+   buffer limit (`MAX_CANON`, 1024 bytes on macOS/Linux), and real Monarch
+   cookie headers are usually longer than that, so the prompt path fails
+   with a confusing auth error for most users. The cookie file has no
+   length limit and survives repo updates.
+
+The script verifies the cookies against the live API before saving them to your system keyring. The cookie file is only read at setup time; the running MCP server uses the keyring session.
 
 #### Option 2: Email and password
 
--- a/login_setup.py
+++ b/login_setup.py
@@ -14,6 +14,7 @@ Supports three auth paths in order of recommendation:
 
 import asyncio
 import getpass
+import os
 import sys
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


robcerda:
The MAX_CANON analysis in #82 is correct and this is the smallest fix that avoids the tty limit entirely. Merging, this closes #82.

PR #89, opened 2026-08-13
Fix update_transaction_rule being silently ignored by the API; add missing rule criteria

CapitalX:
## The bug

`update_transaction_rule` builds its mutation input only from the arguments passed to it. Monarch's `updateTransactionRuleV2` **ignores any input that carries no matching criteria**, so the most natural call — changing just an action — is accepted by the tool and discarded by the API:

```python
update_transaction_rule(rule_id="...", set_category_id="cat_new")
# -> {"success": false, "errors": {"fieldErrors": null, "message": null, "code": null}}
# rule unchanged
```

Verified against the live API. Resending the rule's existing criteria alongside the identical change succeeds. This matters for any agent-driven "audit and clean up my rules" workflow, where changing one action on an existing rule is the main operation.

`update_transaction_rule` now reads the rule first and resends its criteria. Criteria that are *omitted* from the input are preserved by the API, so only the criteria need resending — not every action.

## Also added

Rule features the API accepts but the tools could not reach:

- **`originalStatementCriteria`** — [Monarch's docs](https://help.monarch.com/hc/en-us/articles/360048393372-Transaction-rules) recommend matching on the original statement over the merchant name, since it doesn't change when a merchant is re-identified. It was readable via `get_transaction_rules` but not settable, so a rule using it couldn't be created or reproduced.
- **`categoryIds`** — constrain a rule to transactions already in given categories.
- **Per-value operators** via `merchant_criteria=[{"operator": ..., "value": ...}]`. A single shared operator can't express a rule mixing `contains` and `eq`, which the web UI allows.
- **`amount_operator="between"`**, populating `valueRange`. The operator was already in the docstring but `valueRange` was hardcoded to `None`, so it never worked.
- **The created rule's id**, now selected from the create mutation and returned, so callers don't have to diff the rule list.
- An all-null `PayloadError` is reported as a readable failure instead of being passed through as an empty error object.

All existing arguments keep working unchanged.

## Verification

Exercised against a real Monarch account (throwaway rules, `applyToExistingTransactions: false`, deleted afterwards): create with mixed operators + original-statement + `categoryIds` + a `between` range persists correctly, and an action-only update now applies while preserving every criterion.

`231 passed` — full suite.

## Notes for review

- Three existing update tests mocked a single `gql_call` return value for the whole tool. `update_transaction_rule` now issues a read before its write, so those mocks answer both calls via `side_effect`. `test_update_rule_error` was passing for the wrong reason — its single mock made the read find no rule — and now exercises the error path it describes.
- `delete_transaction_rule` is deliberately untouched: #86 already fixes the `deleted: false` false negative, which I hit independently while testing. This PR is rebased-clean against it (different function).

### DIFF
--- a/src/monarch_mcp_server/tools/rules.py
+++ b/src/monarch_mcp_server/tools/rules.py
@@ -84,12 +84,71 @@ query GetTransactionRules {
     }
     setHideFromReportsAction
     reviewStatusAction
+    sendNotificationAction
+    setLinkToPaydownBudgetAction
+    criteriaOwnerIsJoint
+    criteriaOwnerUserIds
+    criteriaBusinessEntityIds
+    criteriaBusinessEntityIsUnassigned
+    criteriaBusinessEntities {
+      id
+      name
+      __typename
+    }
+    merchantCriteria {
+      operator
+      value
+      __typename
+    }
+    actionSetOwnerIsJoint
+    actionSetOwner {
+      id
+      displayName
+      __typename
+    }
+    actionSetBusinessEntity {
+      id
+      name
+      __typename
+    }
+    actionSetBusinessEntityIsUnassigned
+    linkSavingsGoalAction {
+      id
+      name
+      __typename
+    }
+    needsReviewByUserAction {
+      id
+      displayName
+      __typename
+    }
+    splitTransactionsAction {
+      amountType
+      splitsInfo {
+        categoryId
+        merchantName
+        amount
+        goalId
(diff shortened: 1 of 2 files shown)


robcerda:
Verified the suite at 271 and reviewed the blast radius: the update path aborts if the read fails rather than writing an unscoped rule, and refuses a criteria-less mutation, which is the right call. Merging, this also makes the between operator in #117 reachable.

PR #90, opened 2026-08-13
Add get_goals tool

CapitalX:
## Why

The server exposes `linkGoalAction` on transaction rules, but there is no way to **list** goals — so a client can neither discover a goal id to link a rule to, nor report on goal progress. `goalsV2` is already queried inside `get_cashflow`'s `GetJointPlanningData`, but only as part of budget planning and never surfaced as a tool.

This adds a read-only `get_goals` returning each goal's target / current / starting amounts, planned monthly contribution, archive and completion state, and the accounts allocated to it.

## Notes for reviewers

- **The selection set was derived empirically.** Monarch disables GraphQL introspection for non-admin users and masks unknown-field errors as a generic `"Something went wrong"`, so fields were confirmed one at a time against the live API. Fields verified *absent* are listed in a comment above the query so they aren't speculatively re-added: `targetDate`, `createdAt`, `updatedAt`, `goalBalance`, `balance`, `progress`, `completedPercent`.
- **Debt goals need their own progress formula.** They run from a negative `startingAmount` up to a `targetAmount` of `0`, so the asset formula (`current / target`) reports a nonsensical value. Progress is computed against the amount paid down instead, and returns `None` rather than dividing by zero when the range has zero width. Both paths are covered by tests.
- **Archived goals are included by default.** They still hold their accounts and balances, and a goal that appears "missing" is usually just archived. `include_archived=False` filters them.
- **`name` vs `defaultName`.** `name` is user-editable; `defaultName`/`objective` reflect the template the goal was created from. A goal renamed "End Game" still reports `defaultName: "Retirement"` and `objective: "retirement"`, which is the reliable thing to match on programmatically.

## Verification

Exercised against a real Monarch account (3 goals, one debt and two asset, all archived) plus 7 new unit tests. `229 passed` — full suite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 
 ## 🛠️ Available Tools
 
-All 49 registered tools. Required parameters are listed first, optional ones
+All 53 registered tools. Required parameters are listed first, optional ones
 are marked with a trailing question mark. This table is generated from the
 live tool registry and the functions' signatures, so it does not drift.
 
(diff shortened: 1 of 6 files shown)


robcerda:
Good addition, but this adds four tools including two that write (set_goal_contribution, update_savings_goal) and none are in the README table or the approval list, so it fails the drift tests added in #127. Please add the table rows, list the two mutating ones under the approval section, and add them to MUTATING_TOOLS in read_only.py from #129 so read only mode covers them.

CapitalX:
Thanks @robcerda 

You've done great work on this mcp definitely looking to contribute where I can!

robcerda:
Thorough work, especially documenting that plannedMonthlyContribution does not persist and routing contributions through set_goal_contribution instead. Both mutating tools are gated in read_only.py. Merged manually after resolving conflicts with #91 through #94, plus two type annotations mypy flagged.

PR #91, opened 2026-08-13
Add get_account_sync_health tool

CapitalX:
## Why

A broken institution connection **fails silently**. Monarch keeps serving the last balances it saw and simply stops importing transactions, so budgets, cashflow and spending summaries drift without any error surfacing — including through this MCP server, whose existing tools all return happily against stale data.

Nothing currently exposes the `credentials` root query, so a client has no way to tell whether the numbers it is reading are current. An agent asked "how much did I spend last week?" will confidently answer from a feed that stopped updating.

## What it does

Read-only `get_account_sync_health`, reporting per institution:

- `update_required` — the login needs re-authentication
- `disconnected_at` — the data provider dropped the connection
- `sync_disabled_at` / `sync_disabled_reason`
- `stale_days` — days since the last successful update

Results are split into `needs_attention` and the full `connections` list, so the common case is a single field check.

`stale_after_days` is a parameter rather than a constant because a daily Plaid connection and a slow pension provider have very different normal cadences.

## Found on a real account

An MX-backed connection sitting at `updateRequired: true` and **six days stale**, while seven other institutions were current — exactly the case that stays invisible until you go looking for it.

## Notes for reviewers

- Monarch disables GraphQL introspection for non-admin users and masks unknown-field errors as a generic `"Something went wrong"`, so the selection set was established field-by-field against the live API. Fields verified *absent* are listed in a comment above the query so they aren't speculatively re-added: `lastUpdateAttemptAt`, `dataProviderAccountId`, `dataProviderId`, `status`, `errorType`.
- An unparseable `displayLastUpdatedAt` yields `stale_days: None` rather than raising, so one malformed timestamp can't take down the whole report.

## Verification

Exercised against a real account (8 connections, 1 unhealthy) plus 7 unit tests covering each failure mode, threshold configurability, and the unparseable-timestamp path. `229 passed` — full suite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 
 ## 🛠️ Available Tools
 
-All 49 registered tools. Required parameters are listed first, optional ones
+All 50 registered tools. Required parameters are listed first, optional ones
 are marked with a trailing question mark. This table is generated from the
 live tool registry and the functions' signatures, so it does not drift.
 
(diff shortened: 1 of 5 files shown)


robcerda:
Reads cleanly and get_account_sync_health is genuinely useful for spotting a stale institution. Needs a README table row to pass the drift test added in #127, then I will merge.

robcerda:
Read only, well covered by tests including the unparseable timestamp case, and the README row is in. Merging.

PR #92, opened 2026-08-13
Add reorder_transaction_rule tool

CapitalX:
## Why

Rule order is not cosmetic. Every matching rule runs **in order**, and a later rule **overwrites** an earlier one — so position decides which rule wins when two rules match the same transaction. There was previously no way to change it from this server.

## Why it needs its own mutation

`UpdateTransactionRuleInput` rejects an `order` field (verified). Monarch uses a dedicated mutation:

```graphql
updateTransactionRuleOrderV2(id: ID!, order: Int!)
```

Note it takes **scalar arguments**, not an input object — which is why guessing at an `UpdateTransactionRuleOrderInput` type finds nothing. It returns the full renumbered rule list.

## Behaviour

- Verifies the rule exists first, so a bad id fails without sending the mutation.
- Rejects a negative position without any API call.
- Returns `moved_from` / `moved_to` plus the resulting order of every rule, so the caller can confirm the outcome instead of re-querying.

## Verification

3 unit tests (success, unknown id, negative order). `225 passed` — full suite. Exercised against a live account using throwaway rules, confirming the swap took effect and that the surrounding real rules were renumbered by Monarch and restored on cleanup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 
 ## 🛠️ Available Tools
 
-All 49 registered tools. Required parameters are listed first, optional ones
+All 50 registered tools. Required parameters are listed first, optional ones
 are marked with a trailing question mark. This table is generated from the
 live tool registry and the functions' signatures, so it does not drift.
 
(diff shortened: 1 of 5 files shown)


robcerda:
reorder_transaction_rule is worth having, but this now conflicts with main after #89 reworked the same file. Could you rebase, add a README table row and an entry in MUTATING_TOOLS in read_only.py since it writes, then I will merge.

robcerda:
Rebased cleanly, validates new_order, confirms the rule exists before writing, and returns the resulting order so the caller can verify. Gating in read_only.py and the approval list are both in. Merging.

PR #93, opened 2026-08-13
Add get_debt_paydown tool

CapitalX:
## Why

**Debt paydown is not a goal**, which is easy to get wrong. Monarch's legacy goal system had a `debt` goal type, but the migration to `savingsGoals` **dropped it** — a legacy debt goal comes back from `legacyGoalsMigrationData` with `newGoalId: null` — and replaced it with a plan computed over the debt accounts themselves.

So a client reading only `savingsGoals` sees the user's savings targets and **silently misses their debt entirely**.

## What it does

Read-only `get_debt_paydown`, reporting every debt account (APR, minimum payment, planned payment) alongside the plan's projected interest, projected total and debt-free date.

## The part worth reviewing

Accounts flagged `excludeFromDebtPaydown` are reported **separately**, with the excluded balance totalled against the plan's principal.

This matters because `debtPaydownPlan.currentDebtPrincipal` only covers *included* accounts. Excluding a high-APR card makes the projected debt-free date and total interest look better while the most expensive balance goes untouched — a projection that is technically correct and practically misleading. Surfacing `debt_in_plan` next to `total_debt_across_accounts` and `debt_excluded_from_plan` makes that visible without the caller having to know the trap exists.

On the account this was developed against, three of four cards were excluded — including every card with an APR higher than the single card left in the plan.

## Verification

4 unit tests (plan reporting, excluded-account separation and totalling, method pass-through, error handling). `226 passed` — full suite. Exercised against a live account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 
 ## 🛠️ Available Tools
 
-All 49 registered tools. Required parameters are listed first, optional ones
+All 50 registered tools. Required parameters are listed first, optional ones
 are marked with a trailing question mark. This table is generated from the
 live tool registry and the functions' signatures, so it does not drift.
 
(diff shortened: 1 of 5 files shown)


robcerda:
No objections to get_debt_paydown, it is read only. Needs a README table row to pass the drift test added in #127, then I will merge.

robcerda:
Read only and cleanly tested. Merged manually after resolving the import conflict with #91 and refreshing the README tool counts.

PR #94, opened 2026-08-13
Add monarch_whoami tool with plan-gated capability detection

CapitalX:
## Why

Parts of Monarch's schema are **plan-gated**. Business entities are the clearest case: on a lower tier the fields are absent entirely, so a transaction rule's business-entity fields simply are not there.

Nothing exposed which tier the signed-in account is on, so a client had no way to distinguish *"this account doesn't have that feature"* from *"something went wrong"* — and no way to find a business-entity id when the feature does exist.

## Capabilities are probed, not inferred

The obvious implementation maps entitlement names to features. I did not do that, because it would be guesswork: the names are undocumented, and the mapping is not visible from a single account. This one carries `["premium", "premium_plus_trial"]` and has business entities — not enough to conclude which entitlement grants them.

Instead each capability is a cheap query that either succeeds or does not, which is the same signal the caller would get by trying to use the feature. **A probe that raises is reported as unavailable and never fails the tool** — Monarch masks a missing field as a generic error, so the exception *is* the answer.

`CAPABILITY_PROBES` is a table, so other gated features can be added as contributors confirm them:

```python
CAPABILITY_PROBES = {
    "business_entities": (
        "ProbeBusinessEntities",
        gql("query ProbeBusinessEntities { businessEntities { id name color } }"),
        "businessEntities",
    ),
}
```

Each entry carries the objects the feature needs, so `capabilities.business_entities.items` is also where a business-entity id comes from when setting one on a rule.

## Trial entitlements

Entitlements ending in `_trial` are surfaced separately. A feature available today can **disappear when the trial ends**, taking its schema fields with it — worth knowing before building a rule that depends on one.

## Privacy

`me.birthday` and `me.profilePictureUrl` are deliberately not selected. They are personal data an agent has no use for, and this tool's output tends to end up in transcripts. There is a test asserting they don't appear.

## Verification

7 unit tests, including the gated-off path (probe raises → `available: false`, tool still succeeds) and the PII assertion. `229 passed` — full suite. Exercised against a live account with the feature present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 
 ## 🛠️ Available Tools
 
-All 49 registered tools. Required parameters are listed first, optional ones
+All 50 registered tools. Required parameters are listed first, optional ones
 are marked with a trailing question mark. This table is generated from the
 live tool registry and the functions' signatures, so it does not drift.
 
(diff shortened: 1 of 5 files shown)


robcerda:
monarch_whoami is a reasonable capability probe. Needs a README table row to pass the drift test added in #127, and please confirm the response carries no account identifiers beyond what check_auth_status already exposes.

robcerda:
Thanks for spelling out the PII reasoning, and for excluding birthday and profilePictureUrl. Name and email are the substance of a whoami answer so that is the right call. Merging.

Issue #102, opened 2026-08-13
Launch config: resolve from uv.lock via the monarch-mcp-server console script instead of an unpinned ephemeral env (root cause of #84)


alex-zwingli:
The README's recommended launch command builds a **fresh, unpinned environment on every start** rather than using the environment `uv.lock` already describes:

```
uv run --with 'mcp[cli]' --with-editable /path/to/monarch-mcp-server \
       mcp run /path/to/monarch-mcp-server/src/monarch_mcp_server/server.py
```

`uv.lock` pins `mcp==1.13.1`, but nothing in that command consults the lockfile. `--with 'mcp[cli]'` is an unpinned top-level request, and the project's own floor is `mcp[cli]>=1.10.0`, so the resolver is free to take the newest release.

Since `mcp 2.0.0` is now on PyPI, **it does.** That makes the documented quickstart fail on a cold cache for new users — this is the root cause of #84, which is currently open with only a version-pin workaround.

## Verification

Resolving exactly what that command must satisfy (project deps + unpinned `mcp[cli]`, `requires-python = ">=3.12,<3.14"`) picks v2:

```
$ uv lock   # scratch project mirroring the ephemeral env
Resolved 62 packages in 1.12s
$ grep -A1 'name = "mcp"' uv.lock
name = "mcp"
version = "2.0.0"
```

And v2 no longer ships the module `app.py` imports:

```
$ uv run --isolated --no-project --with 'mcp[cli]==2.0.0' \
    python -c "import importlib; importlib.import_module('mcp.server.fastmcp')"
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

That is the traceback in #84.

Note this is easy to miss when testing locally: if you have ever run `uv sync` in the repo, uv's cache is warm and the same command resolves to the locked 1.13.1, so the server starts fine. The failure only shows up on a machine that has not synced the project — i.e. exactly a new user following the quickstart.

## Proposal

`pyproject.toml` already declares a console script that is never mentioned in the README:

```toml
[project.scripts]
monarch-mcp-server = "monarch_mcp_server.server:main"
```

Documenting `--directory` + that entry point makes the launch command resolve from `uv.lock`:

```json
{
  "mcpServers": {
    "Monarch Money": {
      "command": "uv",
      "args": ["run", "--directory", "/path/to/your/monarch-mcp-server", "monarch-mcp-server"]
    }
  }
}
```

Benefits:

[...]

Issue #84, opened 2026-08-04
Server fails to start: ModuleNotFoundError: No module named 'mcp.server.fastmcp'


yuryu:
As of early August 2026, the server fails to start with `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`. The MCP Python SDK v2 renamed `FastMCP` to `MCPServer` and moved `mcp.server.fastmcp.*` to `mcp.server.mcpserver.*` (see the [v1→v2 migration guide](https://py.sdk.modelcontextprotocol.io/migration/)). Because the recommended launch command uses an unpinned `--with 'mcp[cli]'`, uv now resolves to SDK v2 and the import in `app.py` fails at startup.

## Steps to reproduce
 
1. Configure the server per the README, e.g.:
```
   uv run --with 'mcp[cli]' --with-editable /path/to/monarch-mcp-server mcp run /path/to/monarch-mcp-server/src/monarch_mcp_server/server.py
```
2. Let uv resolve `mcp[cli]` to the latest version (now ≥2.0).
3. Start the server (e.g., via Claude Desktop).
## Expected behavior
 
Server starts and connects.
 
## Actual behavior
 
```
Failed to run server
Traceback (most recent call last):
  File ".../site-packages/mcp/cli/cli.py", line 341, in run
    server = _import_server(file, server_object)
  File ".../site-packages/mcp/cli/cli.py", line 141, in _import_server
    spec.loader.exec_module(module)
  File ".../monarch-mcp-server/src/monarch_mcp_server/server.py", line 10, in <module>
    from monarch_mcp_server.app import mcp, main, app  # noqa: F401
  File ".../monarch-mcp-server/src/monarch_mcp_server/app.py", line 5, in <module>
    from mcp.server.fastmcp import FastMCP
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```
 
The MCP client (Claude Desktop) then reports "Server disconnected."
 
## Environment
 
- macOS (Apple Silicon), Python 3.13 via uv
- `mcp[cli]` unpinned → resolves to SDK v2
- Client: Claude Desktop
## Workaround
 
Pin the SDK to v1 in the launch command:
 
```
--with 'mcp[cli]<2'
```
 
(or pin `mcp<2` in `pyproject.toml`). This restores the previous behavior.

jkmaxwell:
+1, workaround works; thank you.

robcerda:
[...]

PR #104, opened 2026-08-16
Surface Monarch's Flexible budget bucket in get_budgets, with a safe fallback

### LINKED ISSUE #103: get_budgets omits the Flexible bucket amount: query drops monthlyAmountsForFlexExpense / totalsByMonth
## Summary

For accounts on Monarch's **fixed_and_flex** budget system, the "Flexible" section carries a single bucket-level amount covering every category under it. `get_budgets` never reports that amount. It returns the individual categories inside the bucket, but the all-up Flex figure — the number the user actually budgets against — is missing entirely, so spending-vs-budget for Flexible cannot be computed from this tool's output.

There is also no way to *set* it (see "Write side" below).

Confirmed against a live `fixed_and_flex` account; details in "Live verification".

## Root cause

`BUDGET_QUERY` in `src/monarch_mcp_server/tools/budgets.py` (lines 21–55) requests only `budgetData { monthlyAmountsByCategory }`. It omits all three places Monarch exposes bucket-level numbers:

- `monthlyAmountsForFlexExpense` — the Flex bucket's planned/actual/remaining per month
- `monthlyAmountsByCategoryGroup` — group-level rollups
- `totalsByMonth` — including `totalFlexibleExpenses`, `totalFixedExpenses`, `totalNonMonthlyExpenses`

`format_budget_data()` (lines 82–114) matches that shape: it iterates `monthlyAmountsByCategory` only and emits one row per category per month. So even if the fields were returned, nothing would surface them.

For comparison, the installed `monarchmoneycommunity==1.4.0` requests all of them in its own budgets query — `monarchmoney/monarchmoney.py` lines 1264–1331:

```graphql
monthlyAmountsForFlexExpense {
  budgetVariability
  monthlyAmounts {
    month
    plannedCashFlowAmount
    actualAmount
    remainingAmount
    previousMonthRolloverAmount
    rolloverType
  }
}
totalsByMonth {
  month
  totalFlexibleExpenses { plannedAmount actualAmount remainingAmount previousMonthRolloverAmount }
  totalFixedExpenses    { ... }
  totalNonMonthlyExpenses { ... }
}
```

## Why the narrowed query dropped it — and the part that matters for the fix

The comment above `BUDGET_QUERY` (lines 17–20) explains the narrowing:

> The upstream SDK's `get_budgets()` requests category-group fields (e.g. `budgetVariability`/`rolloverPeriod`) that Monarch's current API rejects for some accounts, so it can fail outright. This narrower query asks only for fields the current API still returns.

That is a reasonable workaround, but it removed more than it needed to. **The problematic fields and the flex fields live in different subtrees of the query:**

- Rejected fields are on the top-level `categoryGroups` selection — `groupLevelBudgetingEnabled`, `budgetVariability`, `rolloverPeriod` (upstream lines 1334–1356).
- Flex/bucket data lives under `budgetData` — a sibling selection (upstream lines 1264–1331).

So `monthlyAmountsForFlexExpense` and `totalsByMonth` can be restored **without** reintroducing any of the `categoryGroups` fields that caused the original failure. The current query keeps its narrow `categoryGroups { id name type categories { id name } }` exactly as-is. This is confirmed empirically below.

## Live verification

Run against a live `fixed_and_flex` account, window 2026-08-01 → 2026-08-31, server at `ca6c159`, `monarchmoneycommunity==1.4.0`. Structure only — no amounts.

**1. Baseline, using the shipped `BUDGET_QUERY`:**

```
86 category rows
budgetData keys returned: ['__typename', 'monthlyAmountsByCategory']
categoryGroups: 18
```

`budgetData` comes back with `monthlyAmountsByCategory` and nothing else — the Flex total is absent from the response, not lost during formatting.

**2. The flex/bucket fields are accepted, not rejected.** Adding `monthlyAmountsForFlexExpense` and `totalsByMonth` to the same `budgetData` selection succeeded with **no GraphQL error**:

```
EXTENDED QUERY: ACCEPTED by API

monthlyAmountsForFlexExpense ->
  {'budgetVariability': 'flexible',
   'monthlyAmounts': [len=1, {'month': '2026-08-01',
                              'plannedCashFlowAmount': number(present),
                              'actualAmount': number(present),
                              'remainingAmount': number(present),
                              'previousMonthRolloverAmount': number(present),
                              'rolloverType': 'monthly'}]}

totalsByMonth ->
  [len=1, {'month': '2026-08-01',
           'totalFlexibleExpenses':    {plannedAmount, actualAmount, remainingAmount -- all present},
           'totalFixedExpenses':       {... all present},
           'totalNonMonthlyExpenses':  {... all present}}]

budgetVariability='flexible' months=1 non-zero planned values: 1/1
```

`plannedCashFlowAmount` is present and non-zero — that is the all-up Flexible figure this issue is about.

**3. The Flex bucket is not a category group.** Of the 18 category groups returned, none has a name containing "flex":

```
group names containing 'flex': []
```

**4. The upstream `get_budgets()` that the workaround replaced now succeeds.** Calling `client.get_budgets(start_date, end_date)` directly — the full upstream query, including the `categoryGroups` fields the comment says Monarch rejects — returned without error on this account:

```
RESULT: SUCCEEDED
budgetData keys: ['monthlyAmountsByCategory', 'monthlyAmountsByCategoryGroup',
                  'monthlyAmountsForFlexExpense', 'totalsByMonth']
monthlyAmountsForFlexExpense present: True
  budgetVariability='flexible' months=1 non-zero planned=1
categoryGroups returned: 18
```

So the condition that motivated the narrowed query does not reproduce here. That does not prove it is fixed for every account — the original comment says "for some accounts" — but it does mean the workaround may now be costing flex data for no live benefit, and is worth re-testing before being kept.

## Write side

`set_budget_amount` (lines 143–212) accepts only `category_id` or `category_group_id` and calls `client.set_budget_amount(...)`. Given finding 3 above, `category_group_id` **cannot** reach the Flex bucket by construction, regardless of which ID is passed — the bucket simply is not among the category groups. Setting it requires a separate mutation, which upstream already wraps:

```python
# monarchmoney/monarchmoney.py:2751
async def update_flexible_budget(amount, start_date=None, apply_to_future=False)
#   mutation Common_UpdateFlexBudgetMutation -> updateOrCreateFlexBudgetItem
```

Neither `update_flexible_budget()` nor the adjacent `update_flex_rollover_settings()` (line 2801) is exposed by this server — `grep -ri flex src/` finds only `budget_variability` handling in `tools/categories.py`.

That leaves an odd asymmetry: `update_category(budget_variability="flexible")` can already move a category *into* the Flex bucket, but the bucket's own amount can be neither read nor set.

## Proposal

Since it is unknown whether every account accepts the flex selections, the design below **degrades instead of failing**. The hard requirement: `get_budgets` must not regress for any account that works today.

### 1. Two queries, extended-first with fallback

Keep the existing `BUDGET_QUERY` verbatim as the fallback, and add `BUDGET_QUERY_FLEX` — the same query plus `monthlyAmountsForFlexExpense` and `totalsByMonth`. Prefer this over reverting to `client.get_budgets()`: per finding 4 that works today, but it reintroduces the `categoryGroups` fields blamed in #15, so it carries risk the narrow query does not.

```python
_flex_supported: Optional[bool] = None   # module-level, per-process

async def get_budget_data(client, start_date=None, end_date=None):
    global _flex_supported
    ...
    if _flex_supported is not False:
        try:
            data = await client.gql_call("MCPBudgetDataFlex", BUDGET_QUERY_FLEX, variables)
            _flex_supported = True
            return data
        except Exception as e:
            if _is_schema_rejection(e):
                logger.warning("Flex fields rejected; falling back to narrow query: %s", e)
                _flex_supported = False      # cache ONLY schema rejections
            else:
                raise                        # auth/network: surface, don't mask
    return await client.gql_call("MCPBudgetData", BUDGET_QUERY, variables)
```

**Only latch the negative for schema-level rejections** — match GraphQL validation text (`Cannot query field`, `Unknown field`, `Unknown argument`). This matters: a 401 or a socket timeout must not permanently disable flex for the process, and must not be silently converted into a degraded-but-successful response. Caching the result means an unsupported account pays one failed round-trip per process, not one per call.

### 2. Distinguish three states in the output

Silently omitting the field on failure is indistinguishable from a flex budget of zero — which will lead an LLM to state a confidently wrong number. Make it explicit:

```json
{
  "data": [ "...unchanged per-category rows..." ],
  "flex": {
    "status": "ok | not_configured | unsupported",
    "budget_variability": "flexible",
    "monthly": [ {"month": "...", "planned": 0, "actual": 0, "remaining": 0} ]
  },
  "totals": null
}
```

- `ok` — fields returned and populated
- `not_configured` — query succeeded but flex is null/empty (account not on `fixed_and_flex`)
- `unsupported` — query rejected; narrow fallback was used

`data` keeps its current shape so existing consumers don't break. `format_budget_data()` should traverse with `.get()` throughout and never raise on a missing key.

### 3. Write side

Expose `update_flexible_budget()` as a `set_flexible_budget` tool (or `flex=True` on `set_budget_amount`). It should refuse with a clear message when status is `unsupported` or `not_configured`, rather than surfacing a raw GraphQL exception. Being a mutation, it belongs on the README's approval-required list.

### 4. Tests worth having

Extended path succeeds; schema rejection → fallback used and `status="unsupported"`; flex present but null → `not_configured`; and a 401/timeout propagates as an error rather than silently degrading.

## What is verified vs. still open

**Verified** by reading this repo at `ca6c159` and the installed `monarchmoneycommunity==1.4.0`, and by the live run above: the fields absent from the query, that `format_budget_data()` cannot emit them, that upstream requests them, that the extended selection is accepted and populated on a `fixed_and_flex` account, that the Flex bucket is not a category group, and that no flex read/write path is exposed by the server.

Also verified: `update_flexible_budget()` and `update_flex_rollover_settings()` are present and callable on the installed client (signatures confirmed by introspection; **not** invoked, since they mutate), and the upstream `get_budgets()` succeeds and returns all four `budgetData` keys.

**Still open:**

- Behavior for accounts **not** on `fixed_and_flex`. `monthlyAmountsForFlexExpense` may return null or error there, so the implementation should degrade gracefully rather than fail — the same class of concern that motivated the original narrowing.
- Whether the `categoryGroups` rejection behind #15 still reproduces anywhere. It does not on this account (finding 4); if it reproduces nowhere, the narrowed query could be retired entirely.

## Environment

- Windows 11, Python 3.13.5, uv; server at `ca6c159`, `monarchmoneycommunity==1.4.0`, `mcp==1.13.1`
- Possibly related historical context: #15 ("getBudgets doesn't work"), which appears to be what the narrowed query addressed.

alex-zwingli:
Fixes #103.

Under Monarch's `fixed_and_flex` budget system, the **Flexible** section carries a single amount covering every category beneath it. `get_budgets` never returned that number — only the individual categories inside the bucket — so spending-vs-budget for Flexible could not be computed from the tool's output.

## Why it was missing

`BUDGET_QUERY` requested only `budgetData { monthlyAmountsByCategory }`. The query had been deliberately narrowed to avoid `categoryGroups` fields (`budgetVariability`, `rolloverPeriod`) that Monarch rejects for some accounts.

That narrowing cut wider than it needed to. The rejected fields are on the top-level `categoryGroups` selection; the flex data lives under `budgetData`, a **different subtree**. So the flex fields can be restored with the `categoryGroups` selection left byte-for-byte unchanged — which is what this PR does.

## Approach: extended-first, with a safe fallback

Rather than change `BUDGET_QUERY`, this adds `BUDGET_QUERY_FLEX` alongside it — the same document plus `monthlyAmountsForFlexExpense` and `totalsByMonth`. It is tried first; if Monarch rejects those fields, the original narrow query runs and the tool behaves exactly as it does today.

The important detail is **which failures count as "unsupported"**, and there are two traps here worth calling out.

**1. Monarch's rejection text is generic.** A refused field does not come back with standard GraphQL wording — probing the live API with a deliberately unknown field returns:

```
{'message': 'Something went wrong while processing: None on request_id: None.'}
```

So matching on `Cannot query field` / `Unknown field` would never fire against the real service. Detection keys on the exception **type** instead: `gql` raises `TransportQueryError` when the server answers with GraphQL `errors`, while HTTP failures raise `TransportServerError` and connection problems raise their own types. Text markers are kept only as a backstop for transports that do use the usual wording.

**2. A failure that hits both queries isn't evidence about flex.** The `unsupported` latch is set only *after* the narrow query succeeds:

```python
except Exception as exc:
    if not _is_query_rejection(exc):
        raise                    # HTTP / connection -> surface it
    rejection = exc              # not latched yet

data = await client.gql_call(... BUDGET_QUERY ...)   # raises -> propagates
if rejection is not None:
    _flex_supported = False      # only now: flex failed, narrow worked
```

An expired session refuses *both* queries. Latching on the first failure would disable flex for the rest of the process because of an auth problem; this way it propagates and flex stays un-probed. An account that genuinely lacks the fields still pays only one failed round-trip per process rather than one per call.

Also confirmed while designing this: `@include(if: false)` does **not** let a single document carry optional flex fields — GraphQL validates the whole document before directives apply, and the live API rejects it. Two documents are genuinely required, so both render from one template to keep the `categoryGroups` selection structurally identical.

## Output shape

`get_budgets` now returns an object instead of a bare list:

```json
{
  "tool": "get_budgets",
  "args": {"start_date": null, "end_date": null},
  "data":   [ "... unchanged per-category rows ..." ],
  "budget_system": "fixed_and_flex",
  "flex":   {"status": "ok", "budget_variability": "flexible", "monthly": [ "..." ]},
  "groups": [ {"id": "...", "name": "...", "planned": 0, "actual": 0, "month": "..."} ],
  "goals":  [ {"id": "...", "name": "...", "archived": false,
               "planned_contributions": [], "actual_contributions": []} ],
  "totals": [ {"month": "...", "income": {}, "expenses": {},
               "flexible": {}, "fixed": {}, "non_monthly": {}} ]
}
```

Each per-category row in `data` also gains `set_aside`, `rollover` and `rollover_type`.

## Two further gaps above the per-category level

Auditing what else Monarch returns above individual categories turned up two more omissions, both in data the API already sends:

- **`totalsByMonth.totalIncome` and `.totalExpenses` were never requested.** Income was missing from the tool altogether — only the three expense sub-buckets were mapped. On a live account both come back fully populated.
- **`monthlyAmountsByCategoryGroup` was never requested.** Groups budgeted at the group level rather than per category hold their amount there; it now surfaces as `groups`.

Both ride the same extended selection and therefore the same fallback, so an account that refuses these fields still gets every category row with `flex`/`groups`/`totals` marked unavailable.

`groupLevelBudgetingEnabled` is deliberately **not** requested — it sits on `categoryGroups`, which this PR keeps narrow. Group names are resolved from the existing narrow selection instead.

One asymmetry worth recording so it isn't mistaken for a further gap: **Flexible is the only *pooled* bucket.** Fixed and Non-Monthly are budgeted per category, so their per-category rows in `data` were already complete and their totals are plain sums — which is why Monarch exposes a dedicated `monthlyAmountsForFlexExpense` and no fixed/non-monthly equivalent. That's noted in the tool docstring.

## Rollover: per-category numbers that did not add up

A further pass over the per-category fields found the most consequential omission. Monarch returns `previousMonthRolloverAmount` and `rolloverType` per category; neither was requested. For any category with rollover enabled, `planned - actual` therefore did **not** equal `remaining`, and nothing in the response explained the difference — it just looked like bad data.

On a live account, 5 of 86 rows were affected. With rollover included:

```
rows: 86
  reconcile only once rollover is included: 5
  still unexplained:                        0
```

Also in this pass:

- **`plannedSetAsideAmount` was already being requested and then silently dropped** during formatting. Now mapped as `set_aside`.
- **`budgetSystem`** is now requested and returned as top-level `budget_system` (e.g. `"fixed_and_flex"`), so a caller can distinguish "this account does not use flex budgeting" from "the flex bucket is empty" rather than inferring it from an empty result.

These go into the extended query only, through new template slots. The narrow fallback stays byte-for-byte the document already proven to work, and `test_fallback_query_requests_nothing_beyond_the_proven_set` asserts none of the extended fields leak into it — that query is the safety net, so it must never become the thing that breaks.

`flex.status` is deliberately explicit, because silently omitting the field is indistinguishable from a flex budget of zero — and an LLM reading it that way would state a confidently wrong number:

| status | meaning |
|---|---|
| `ok` | bucket present and populated |
| `not_configured` | query succeeded, but this account has no flex bucket |
| `unsupported` | Monarch rejected the flex fields; fallback query ran |

**This is the one breaking change:** callers reading the old top-level list now read `data`. It matches the envelope `get_transactions` already returns, and the README is updated accordingly.

## Write side

Added `set_flexible_budget`, wrapping the client's existing `update_flexible_budget()`. This is needed because the Flex bucket is **not** a category group — of the 18 category groups returned on my account, none corresponds to it — so `set_budget_amount(category_group_id=...)` cannot reach it by construction. It refuses early with a clear message when flex is known-unsupported, instead of surfacing a raw GraphQL exception, and is added to the README's approval-required list since it mutates the ledger.

## Testing

**Unit: 69 budget tests, up from 6.** Full suite **284 passed**.

Coverage spans the flex-present / `not_configured` / rejection→fallback paths; that a transient rejection recovers rather than sticking; that `TransportServerError` propagates instead of degrading; that a failure hitting **both** queries surfaces as an error; rejection detection against the real generic Monarch message rather than idealized GraphQL wording; both list and bare-object shapes for `monthlyAmountsForFlexExpense` and the refusal to merge multiple buckets; explicit JSON nulls at every level; the income/expense marker; group-level budgets; paired-date validation in both directions at both the tool and helper level; and every branch of the two mutating tools including their confirming-payload guards.

**These were validated by mutation, not by inspection.** Each of the following was applied to the implementation and confirmed to fail a test: swapping the two query documents (both as constants and at the call sites), returning `used_flex_query=True` on the fallback path, making `json_error` return a success-looking payload, hardcoding a wrong `current_month_range`, dropping `budget_system=` from the destructive client call, dropping the `category_type` key, removing the null-guards, and disabling either confirming-payload guard. Several of those mutations passed an earlier version of this test suite, which is why the tests were rewritten.

`test_fallback_query_requests_nothing_beyond_the_proven_set` guards the regression this PR is most careful about: the narrow document is the safety net, so no extended field may leak into it.

Full suite: **284 passed**. One unrelated failure, `test_secure_session.py::TestGetAuthenticatedClient::test_no_session_returns_none`, is **pre-existing on `main`** — confirmed by stashing these changes and re-running.

Worth flagging separately, since it will not reproduce in CI: that test fails only on a machine where someone has actually authenticated. `_TOKEN_DIR` resolves from `Path.home()` at import, and the fixture fakes the keyring but not the file fallback, so the test reads the developer's real `~/.monarch-mcp-server/token` and gets a live session where it expects `None`. Running it with `HOME` pointed at a temp dir passes. Happy to fix that in a separate PR — it's untouched here.

**Live:** verified against a real `fixed_and_flex` account throughout. `get_budgets` returns 86 category rows with `flex.status="ok"`, a populated bucket, group roll-ups, goals and all five `totals`; every row reconciles once rollover is included (5 previously could not, 0 remain); an explicit date range resolves to the right month; and forcing the fallback still returns all 86 rows with `flex.status="unsupported"` and the extended keys null — the no-regression guarantee.

Neither mutating tool was executed against the live account, since both write to a real ledger. They are covered by unit tests only, which is the main limitation of this PR's verification.

**Reviewed by four independent audits** (correctness, test quality, GraphQL mapping vs the vendored client, docs consistency) over three rounds. Everything they raised in scope is fixed; findings outside this PR's scope were filed separately as #105, #106 and #107 rather than folded in here.

## Goals

`goalsV2` comes back from the same query and was not requested either. Goal contributions are precisely what `plannedSetAsideAmount` refers to, so a "what is spoken for this month" answer that ignores them understates the plan. Surfaced as `goals`, with `planned_contributions` and `actual_contributions` per month.

Archived and completed goals are included but **flagged** rather than filtered, so a caller can exclude them deliberately instead of never learning they existed. Requested without upstream's `@include(if: $useV2Goals)` guard, since the extended document already falls back wholesale if Monarch refuses any part of it.

## `update_flex_rollover_settings`, with its defaults removed

Also exposed, since it is the fix for a Flex bucket that has accumulated a large negative rollover over many months.

It needs care. The client signature is `(rollover_start_month=None, rollover_starting_balance=0.0, rollover_enabled=True)`, and its own docstring offers this as the usage example:

> Example (reset flex rollover to $0 starting this month): `await mm.update_flex_rollover_settings()`

So a zero-argument call silently discards accumulated rollover. Inheriting those defaults would hand an LLM a no-argument reset button, so **both destructive arguments are required** at the tool boundary, the docstring states plainly what is discarded and asks that the current rollover be reported first, and it is on the README's approval-required list. A test asserts the two arguments stay required.

## Deliberately left out

- **`get_cashflow_summary()`** — redundant. Verified against the live API that the existing `get_cashflow` tool already returns the same block (`sumIncome`, `sumExpense`, `savings`, `savingsRate`); this method returns that and nothing else. A second tool returning a strict subset of an existing one is surface an agent has to choose between for no gain.
- **`reset_budget()`** — clears a whole month of planned amounts with no undo. Different blast radius from the scoped, re-settable mutations here; it deserves its own discussion before being handed to an LLM.
- **`reset_budget()`** is the one upstream budget capability deliberately left unexposed, for the reason above.

Happy to follow up on any of these separately.

## If you'd rather this were smaller

This is a large PR because the pieces share one query, one fallback path and one response shape, and splitting them means a rebase chain. But it is deliberately structured so it *can* be split, and I'm happy to do that instead of asking you to review it whole. Each commit is self-contained and the natural cut lines are:

| Slice | Contents | Breaking |
|---|---|---|
| A | Extended query + fallback machinery + per-category `rollover` / `rollover_type` / `set_aside` — return stays a bare list, rows just gain keys | **no** |
| B | The `{data, flex, groups, goals, totals, budget_system}` envelope | **yes** |
| C | Write tools: `set_flexible_budget`, `update_flex_rollover_settings` | no |

The one I'd most suggest, if you want only one: **pull A out from under B.** The rollover fix is the highest-value part — without it, `planned - actual` silently fails to equal `remaining` for any rollover category — and it needs no shape change at all. Landing A alone fixes real wrong numbers today, and leaves the list→object question to be settled separately on its own merits.

C depends only on A. B is the only piece anyone should have to argue about.

Two unrelated follow-ups I'd send separately rather than smuggle in here:

- The test suite reads the developer's real `~/.monarch-mcp-server/token` (see Testing below) — a small fixture fix.
- The README's Claude Desktop launch command resolves `mcp[cli]` unpinned, which now pulls SDK v2 and breaks a cold-cache install (that's #84; I opened #102 with the details).

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 - **Split Transaction**: Divide a single transaction into multiple parts with different categories or merchants
 
 ### 💵 Budget Management
-- **Get Budgets**: Access budget information including spent amounts and remaining balances by category, plus the all-up **Flexible** bucket and per-month fixed/flexible/non-monthly totals
+- **Get Budgets**: Budget information by category (including rollover and set-aside), the all-up **Flexible** bucket, category-group roll-ups, savings-goal contributions, the account's budget system, and per-month income / expense / fixed / flexible / non-monthly totals
 - **Set Budget Amount**: Create or modify budget amounts for any category or category group
 - **Set Flexible Budget**: Set the bucket-level Flexible amount used by Monarch's "fixed_and_flex" budget system
 - **Update Flex Rollover Settings**: Start a fresh Flex rollover period, for buckets that have accumulated a large negative rollover
@@ -261,7 +261,7 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 | `check_auth_status` | Check authentication status | None |
 | `get_accounts` | Get all financial accounts | None |
 | `get_transactions` | Get transactions with filtering and reconciliation fields | `limit`, `offset`, `start_date`, `end_date`, `account_id`, `account_ids`, `category_ids`, `category_group_ids`, `tag_ids`, `search`, `wide_search`, `search_scan_limit`, `has_notes`, `is_split`, `is_recurring` |
-| `get_budgets` | Get budget information, including the Flexible bucket | `start_date`, `end_date` |
+| `get_budgets` | Get budget information, including the Flexible bucket | `start_date`, `end_date` (pass both or neither) |
 | `set_budget_amount` | Set budget for a category | `amount`, `category_id`, `category_group_id`, `start_date`, `apply_to_future` |
 | `set_flexible_budget` | Set the all-up Flexible bucket amount | `amount`, `start_date`, `apply_to_future` |
 | `update_flex_rollover_settings` | Start a new Flex rollover period (**discards accumulated rollover**) | `rollover_start_month`, `rollover_starting_balance`, `rollover_enabled` |
@@ -331,11 +331,29 @@ Use get_budgets to show my current budget status
 - `not_configured` — the account has no Flexible bucket (e.g. it is not on flex budgeting)
 - `unsupported` — Monarch rejected the extended fields, so the narrower fallback query ran
 
-A status other than `ok` means **no amount is available — it does not mean zero**. When it is `unsupported`, `groups`, `goals`, `totals`, `budget_system` and each row's `rollover` / `budget_variability` are all `null` for the same reason.
+A status other than `ok` means **no amount is available — it does not mean zero**. When it is `unsupported`, `groups`, `goals`, `totals`, `budget_system` and each row's `rollover` / `rollover_type` / `budget_variability` are all `null` for the same reason.
 
 For `groups`, `goals` and `totals`, `null` means "could not be fetched" while `[]` means "asked, and there were none".
 
-Per-row, `planned - actual` equals `remaining` only when `rollover` is zero; for rollover categories the carried balance accounts for the difference. In `groups`, add a group's `planned` to its categories' totals only when `group_level_budgeting` is false — when true the group holds the budget itself.
+Per-row, `planned - actual` equals `remaining` only when `rollover` is zero; for rollover categories the carried balance accounts for the difference.
+
+**Never add a group's `planned` to its categories' `planned`** — that double-counts either way. `group_level_budgeting` tells you which level holds the real budget: when `true` the group does and its categories carry none; when `false` the group row is only the roll-up of those same categories.
+
+Full response shape:
+
+| Key | Contents |
+|---|---|
+| `tool`, `args` | echo of the call |
+| `budget_system` | e.g. `"fixed_and_flex"`, or `null` |
+| `data[]` | `id`, `name`, `planned`, `actual`, `remaining`, `set_aside`, `rollover`, `rollover_type`, `category_group`, `category_type`, `budget_variability`, `month` |
+| `flex` | `status`, `budget_variability`, `monthly[]` (`month`, `planned`, `actual`, `remaining`, `rollover`, `rollover_type`) |
+| `groups[]` | `id`, `name`, `planned`, `actual`, `remaining`, `rollover`, `rollover_type`, `category_type`, `group_level_budgeting`, `month` |
+| `goals[]` | `id`, `name`, `priority`, `archived`, `completed`, `planned_contributions[]`, `actual_contributions[]` |
+| `totals[]` | `month`, plus `income` / `expenses` / `flexible` / `fixed` / `non_monthly`, each `planned`, `actual`, `remaining`, `rollover` |
+
+Goal contributions are a **separate** quantity from a category's `set_aside`; adding them together double-counts.
+
+`start_date` and `end_date` must be passed **together or not at all** — supplying only one is rejected rather than filled in, because completing the missing side from today can invert the range and return an empty result that reads as "no budget".
 
 ### Set the Flexible Bucket Amount
 ```
--- a/README.md
+++ b/README.md
@@ -323,7 +323,7 @@ Use get_budgets to show my current budget status
 **Two things that will produce wrong numbers if ignored:**
 
 1. **Income and expense amounts are both positive magnitudes.** The sign does not distinguish them — negatives appear only for contra entries. Summing `planned` across all `data` rows adds income to spending. Filter on `category_type` (`income` / `expense` / `transfer`) first.
-2. **Under flex budgeting most category rows carry no standalone budget.** Rows whose `budget_variability` is `flexible` are pooled into the single Flexible bucket, so their individual `planned` values are typically 0 and the real number lives in `flex`. On a representative account 57 of 86 rows were pooled this way; summing the rows gave `7,901` against a true planned expense total of `11,885`.
+2. **Under flex budgeting most category rows carry no standalone budget.** Rows whose `budget_variability` is `flexible` are pooled into the single Flexible bucket, so their individual `planned` values are typically 0 and the real number lives in `flex`. On a flex-budgeting account the majority of rows are usually pooled this way, so adding up the category rows understates planned spending by the whole size of the Flexible bucket. Compare against `totals[].expenses` rather than trusting a row sum.
 
 `flex.status` is one of:
 
@@ -357,7 +357,7 @@ Goal contributions are a **separate** quantity from a category's `set_aside`; ad
 
 ### Set the Flexible Bucket Amount
 ```
-Set my flexible budget to $4,000 for this month using set_flexible_budget
+Set my flexible budget to $1,500 for this month using set_flexible_budget
 ```
 
 ### Set a Budget Amount
--- a/README.md
+++ b/README.md
@@ -318,15 +318,24 @@ Show me my last 50 transactions using get_transactions with limit 50
 Use get_budgets to show my current budget status
 ```
 
-`get_budgets` returns a JSON object with `data` (one row per category per month), `flex`, and `totals`. On Monarch's "fixed_and_flex" budget system the Flexible section carries a **single amount covering every category beneath it**, and that number lives in `flex`, not in the per-category rows — comparing spending against the category rows alone will understate the Flexible budget.
+`get_budgets` returns a JSON object with `budget_system`, `data` (one row per category per month), `flex`, `groups`, `goals` and `totals`.
+
+**Two things that will produce wrong numbers if ignored:**
+
+1. **Income and expense amounts are both positive magnitudes.** The sign does not distinguish them — negatives appear only for contra entries. Summing `planned` across all `data` rows adds income to spending. Filter on `category_type` (`income` / `expense` / `transfer`) first.
+2. **Under flex budgeting most category rows carry no standalone budget.** Rows whose `budget_variability` is `flexible` are pooled into the single Flexible bucket, so their individual `planned` values are typically 0 and the real number lives in `flex`. On a representative account 57 of 86 rows were pooled this way; summing the rows gave `7,901` against a true planned expense total of `11,885`.
 
 `flex.status` is one of:
 
 - `ok` — a Flexible bucket amount was returned
 - `not_configured` — the account has no Flexible bucket (e.g. it is not on flex budgeting)
-- `unsupported` — Monarch rejected the flex fields, so the narrower fallback query ran
+- `unsupported` — Monarch rejected the extended fields, so the narrower fallback query ran
 
-A status other than `ok` means **no amount is available — it does not mean zero**. `totals` carries per-month `flexible` / `fixed` / `non_monthly` totals, or `null` when unavailable.
+A status other than `ok` means **no amount is available — it does not mean zero**. When it is `unsupported`, `groups`, `goals`, `totals`, `budget_system` and each row's `rollover` / `budget_variability` are all `null` for the same reason.
+
+For `groups`, `goals` and `totals`, `null` means "could not be fetched" while `[]` means "asked, and there were none".
+
+Per-row, `planned - actual` equals `remaining` only when `rollover` is zero; for rollover categories the carried balance accounts for the difference. In `groups`, add a group's `planned` to its categories' totals only when `group_level_budgeting` is false — when true the group holds the budget itself.
 
 ### Set the Flexible Bucket Amount
 ```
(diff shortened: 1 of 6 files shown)


alex-zwingli:
@robcerda This PR is ready for your review. It fixes some major gaps that prevent users from effectively budgeting with it. Thanks for getting started on this MCP server!

Let me know if you'd also like help maintaining this repo.

robcerda:
Surfacing the Flexible bucket is clearly right and I like that the destructive rollover arguments are required rather than defaulted, which is the pattern I followed in #125 for update_category. It conflicts with main now, so please rebase and add README rows for any new tools.

PR #109, opened 2026-08-28
Expose Institution(Account) Owner

bradleyseanf:
Hi,

I am the owner of the downstream `monarchmoneycommunity` in `1.5.2` I recently exposed the feature where Monarch allows you to have specific owners assigned to different financial institutions and you can have I believe it is up to 5 users on one single Monarch Account with their users. 

Sometimes -> Especially with an MCP you want to ask the question "Look at all of XYZ's accounts" this opens the door to that. If you have multiple users on the Monarch account.

So this does that:

1.) Expose the ownedByUser entity
2.) Updates mmc to 1.5.2 so it is usable by this package

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
 requires-python = ">=3.12,<3.14"
 dependencies = [
     "mcp[cli]>=1.10.0",
-    "monarchmoneycommunity>=1.4.0",
+    "monarchmoneycommunity>=1.5.2",
     "gql>=3.4",
     "keyring>=24.0.0",
     "python-dotenv>=1.0.0",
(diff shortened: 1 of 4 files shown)


robcerda:
Useful capability and I want it, but as written `account.get("ownedByUser").get("displayName")` raises AttributeError whenever the key is absent or null, which fails three existing tests in test_accounts.py and would break get_accounts entirely for anyone whose payload lacks it. Could you switch to `(account.get("ownedByUser") or {}).get("displayName")`, use snake_case for the key to match the other fields, and add a test covering the null case? Happy to merge after that.

bradleyseanf:
Good point, just added that. Although that does beg the question, if you are a single user on a Monarch account, do you get the `ownedByUser` attribute at all?

robcerda:
Null check and test look right, and I verified monarchmoneycommunity 1.5.2 does select ownedByUser in the accounts query, so the field is actually populated rather than always null. Also checked 1.5.2 keeps every API this repo uses, including redirect_posted on get_transaction_details. Merging.

PR #111, opened 2026-08-29
fix: bound get_cashflow and get_spending_summary response sizes

dwolner:
## The bug

`get_cashflow` returns upstream's raw `Common_GetCashFlowEntityAggregates` payload verbatim. That payload carries a `__typename` on every node, a full Cloudinary logo URL per merchant, and an unbounded merchant list.

On a real account, **one month came to 97,422 characters** and was rejected by the MCP response limit — the tool never returned data at all:

```
Error: result (97,422 characters) exceeds maximum allowed tokens.
```

`get_spending_summary` has the same defect, just less visible: it already strips the GraphQL noise, so a single month fits. Wider ranges don't — a full year is **173,439 characters** and all time is **609,741**.

## The fix

Flatten the aggregates, drop the GraphQL noise, and rank each breakdown by absolute cash flow. Rows are trimmed **only when the response would actually overflow**, stepping a cap down a ladder until it fits — stripping the noise alone gets most ranges under the limit on its own, so a typical month is returned complete.

Every breakdown is now `{total, returned, truncated, rows}`, so a cap is never silent. `limit=N` overrides the automatic behaviour; `limit=0` disables trimming entirely.

The shared machinery (`strip_typenames`, `breakdown`, `render_within_budget`) lives in `helpers.py` and both tools use it, so they share one implementation and one response shape.

### Measured against a live account

| Range | Before | After | Cap applied | Rows returned |
|---|---|---|---|---|
| One month | 97,422 ❌ rejected | 30,783 | none | 41/41 categories, 15/15 groups, 126/126 merchants |
| Full year | 173,439 ❌ | 33,371 | 100 | 70/70, 16/16, 100/918 |
| All time | 609,741 ❌ | 36,115 | 100 | 82/82, 16/16, 100/3566 |

### Merchant ranking

Merchants rank on `|income| + |expense|`, not expense alone. `get_spending_summary` previously sorted by expense only, which drops the largest single line of a period whenever that line is income rather than an expense — common for payroll and transfers, and the case in the account this was tested against. There's a test pinning this in both tools.

## Breaking change

`get_spending_summary`'s breakdowns were bare lists and are now wrapped in the counts envelope. The counts are what make a trim visible; the alternative is a tool that silently returns a fraction of the data.

## Tests

239 pass (222 on `main`). The old `get_cashflow` tests passed against a hand-written `{"cashflow": {"income": 5000}}` mock that upstream never returns — which is exactly why this shipped. The fixture now carries a realistic raw payload, and there are regression tests for the size ceiling, the truncation counts, and the ranking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,7 @@ Once authenticated, use these tools directly in Claude Desktop or Claude Code:
 | `get_transactions` | Get transactions with filtering and reconciliation fields | `limit`, `offset`, `start_date`, `end_date`, `account_id`, `account_ids`, `category_ids`, `category_group_ids`, `tag_ids`, `search`, `wide_search`, `search_scan_limit`, `has_notes`, `is_split`, `is_recurring` |
 | `get_budgets` | Get budget information | `start_date`, `end_date` |
 | `set_budget_amount` | Set budget for a category | `amount`, `category_id`, `category_group_id`, `start_date`, `apply_to_future` |
-| `get_cashflow` | Get cashflow analysis | `start_date`, `end_date` |
+| `get_cashflow` | Get cashflow analysis | `start_date`, `end_date`, `limit` |
 | `get_net_worth` | Get net worth history | `start_date`, `end_date`, `account_type` |
 | `get_account_balance_history` | Get account balance history | `account_id` |
 | `get_net_worth_by_account_type` | Get net worth by account type | `start_date`, `timeframe` |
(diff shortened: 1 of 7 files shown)


robcerda:
Bounding these responses is the right call, the aggregate queries genuinely overflow. It now conflicts with main after #126 moved get_transactions_needing_review onto the shared envelope, so please rebase, and consider reusing tool_response_envelope for the breakdowns rather than the parallel breakdown helper.

PR #113, opened 2026-09-01
Fix keyring save failure on Windows by chunking oversized session blobs

blazemalan:
## Problem

On Windows, saving a cookie-mode session fails with:

```
keyring save error (1783, 'CredWrite', 'The stub received bad data')
```

and the session silently falls back to **plaintext file storage** at `~/.monarch-mcp-server/token` — while the logs make it look like secure storage is working.

Root cause: Windows Credential Manager caps credential blobs at ~2560 bytes (`CRED_MAX_CREDENTIAL_BLOB_SIZE`), and keyring's WinVault backend stores passwords as UTF-16 (2 bytes/char), so anything over ~1280 characters is rejected. The cookie-mode JSON blob introduced in #73 (token + device_uuid + cookies including `cf_clearance`) essentially always exceeds that, so on Windows cookie auth *always* degrades to the plaintext fallback.

## Fix

`secure_session.py` now chunks oversized blobs across multiple keyring entries:

- Blobs ≤ 1024 chars keep the exact single-entry format at `monarch-token` — no change for existing installs.
- Larger blobs are split into 1024-char chunks under `monarch-token-chunk-0..N-1`, with a small JSON index (`{"__monarch_chunks__": N}`) at `monarch-token`.
- Chunks are written before the index, so a crash mid-save can never leave an index pointing at missing chunks.
- Saves and `delete_token` sweep stale chunk entries, so shrinking sessions leave no orphans.
- A chunked entry with a missing chunk is treated as "no session" (falls through to the file fallback) rather than returning truncated JSON.
- Backward compatible with all existing formats: single-entry JSON, legacy bare-token strings, and file-fallback sessions load unchanged.

Chunking was chosen over zlib+base64 compression because `cf_clearance` is high-entropy data that barely compresses — compression cannot guarantee fitting under the limit.

## Probe hardening

`_keyring_available()` round-tripped a 1-byte sentinel, which passes on Windows even though real saves fail. The probe now round-trips a full chunk-sized (1024-char) payload, so a backend that accepts tiny writes but rejects real ones is no longer trusted.

## Tests

New `TestChunkedKeyringStorage` class covering:

- oversized cookie-blob save/load round-trip
- a simulated Windows backend that raises error 1783 on writes over 1280 chars — proving the save now stays in the keyring with no plaintext fallback
- stale-chunk cleanup when sessions shrink, chunked deletion, missing-chunk handling
- format stability for small blobs (no chunk marker leaks into normal sessions)

Also sandboxed the file-fallback path in the shared test fixture: tests previously read the real `~/.monarch-mcp-server/token`, so `test_no_session_returns_none` failed on any machine where the fallback had ever triggered.

Full suite: **231 passed**.

### DIFF
--- a/src/monarch_mcp_server/secure_session.py
+++ b/src/monarch_mcp_server/secure_session.py
@@ -24,12 +24,27 @@ logger = logging.getLogger(__name__)
 KEYRING_SERVICE = "com.mcp.monarch-mcp-server"
 KEYRING_USERNAME = "monarch-token"
 
+# Windows Credential Manager rejects credential blobs over ~2560 bytes
+# (CRED_MAX_CREDENTIAL_BLOB_SIZE) with (1783, 'CredWrite', 'The stub
+# received bad data'), and keyring's WinVault backend stores passwords as
+# UTF-16 (2 bytes per char). A cookie-mode session blob (token +
+# device_uuid + cookies incl. cf_clearance) easily exceeds that, so blobs
+# longer than this many characters are split across multiple entries:
+# an index entry at KEYRING_USERNAME plus monarch-token-chunk-0..N-1.
+_KEYRING_CHUNK_SIZE = 1024
+_CHUNK_USERNAME_PREFIX = KEYRING_USERNAME + "-chunk-"
+_CHUNK_MARKER = "__monarch_chunks__"
+# Safety cap when sweeping chunk entries so a misbehaving backend that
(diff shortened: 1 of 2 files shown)


robcerda:
Good catch that a one byte probe proves nothing about a real cookie blob on Windows Credential Manager. Verified the 1024 char probe still passes against the real macOS Keychain and the full suite is green at 237, merging. This closes #80.

PR #130, opened 2026-09-06
feat(accounts): add update_account tool

george-viaud:
## What

Adds an `update_account` MCP tool. The account surface was read-only apart from `refresh_accounts` and `upload_account_balance_history`, so renaming an account, excluding one from net worth, or hiding its transactions all meant leaving the agent and using the web UI.

`MonarchMoney.update_account` already exists in the underlying library and `Common_UpdateAccount` is already defined — this only wires them up.

## Why

Two cases hit in ordinary use:

**Generic aggregator names.** Institutions supply names like `CREDIT CARD (...1234)`. When several cards come from one issuer, the display name can't tell them apart, and a report read months later is unreadable. Renaming is a one-line fix that currently requires the UI.

**Custodial accounts.** UTMA/UGMA accounts are the child's property, not the custodian's, so counting them overstates net worth. `include_in_net_worth=False` is the correct treatment and there was no way to set it from a tool.

## Design notes

- **Only fields the caller passes are forwarded**, so an unrelated setting can't be reset as a side effect.
- The guard is `is not None`, not a falsy check. `include_in_net_worth=False` is the entire point of the flag, and a balance of `0.00` is a legitimate correction for a paid-off account — a falsy check would silently drop both.
- **Registered in `MUTATING_TOOLS`**, so read-only deployments never see it.
- **Supports `dry_run=True`**, reporting current vs proposed values without writing, matching the convention in `upload_account_balance_history` and `update_category`.
- Errors follow the existing `json_error(tool_name, exc)` shape, and Monarch payload errors are surfaced rather than swallowed.

## Tests

Nine new tests in `tests/test_accounts.py` covering the write path, field filtering, the `False` and `0.0` cases, the empty-update error, both `dry_run` branches, and Monarch payload errors.

`README.md` tool table updated so `test_readme_tool_reference` stays green.

Full suite: **325 passed**. Verified `update_account` is absent from the 25 tools registered under `MONARCH_MCP_READ_ONLY=1` and present in the 50 registered normally.

## Not included

The household/member owner field isn't exposed by the library's `update_account`, so this doesn't cover it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,7 @@ live tool registry and the functions' signatures, so it does not drift.
 | `monarch_login` | Sign in via a secure form in the client UI | None |
 | `monarch_login_with_token` | Sign in with a browser copied session token | None |
 | `monarch_logout` | Clear the stored session and drop the cached client | None |
+| `update_account` | Update an account's name, balance, type or visibility settings | `account_id`, `name`?, `balance`?, `account_type`?, `account_sub_type`?, `include_in_net_worth`?, `hide_from_summary_list`?, `hide_transactions_from_reports`?, `dry_run`? |
 | `refresh_accounts` | Request account data refresh from financial institutions | `account_ids`? |
 | `review_recurring_stream` | Set the review status of a recurring transaction stream | `stream_id`, `review_status` |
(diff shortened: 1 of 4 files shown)


robcerda:
Well scoped and the is not None guard reasoning is right. Merged with three adjustments: added the missing server.py re-export, switched the payload error path to the shared payload_errors helper from #125 so rejections use the same shape as the other mutators, and regenerated the README table after the conflict with #90 through #94.

Issue #100, opened 2026-08-13
Make dependency bumps reviewable: sdist diff artifact on bump PRs + PyPI-vs-tag verification


alex-zwingli:
## Context

Builds on #95. Scanners catch *known*-bad versions; this issue is about catching a **novel** compromise of `monarchmoneycommunity` (or another dependency) at the moment it tries to enter: the version-bump PR.

`monarchmoneycommunity` deserves special attention — it's the code that holds the Monarch session token in request headers, and this repo already reaches into its private internals (`_headers`, `_cookies`, `_auth_mode`, endpoint patching in `monarch_auth.py`), so behavioral review of upstream changes is warranted anyway.

## Proposed checks

**1. Dependabot (or Renovate) for dependency bumps** — so bumps arrive as PRs with a consistent shape the checks below can hook.

**2. Sdist diff artifact on bump PRs.**
A workflow that detects a `uv.lock` change to a watched package, downloads the old and new sdists from PyPI, and attaches a unified diff as a PR artifact (or sticky comment for small diffs). A human then reviews the *actual upstream code change*, not just a version string. This is the single check most likely to catch a novel compromise.

**3. PyPI-vs-tag verification.**
Compare the new PyPI sdist against the corresponding upstream GitHub tag and fail if the PyPI artifact contains files or changes the repo tag doesn't — the classic "tarball differs from tag" attack (seen in xz-utils). If upstream ever publishes PEP 740 attestations / Trusted Publishing provenance, verify that instead.

## Interaction with the egress canary

These checks are complementary to the egress canary in the supply-chain CI issue: the diff review catches malicious *code* at the door; the canary catches malicious *behavior* that review missed; the runtime guard (#96) catches whatever survives both.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code) after a source review of the repo.

robcerda:
Partially done in #132. uv lock --check runs in CI, which catches the case that prompted this: a manifest bump that leaves the lock stale and silently disables hash verification. Dependabot now proposes both action SHA and Python dependency bumps. Not done: the sdist diff artifact and PyPI versus tag verification, which are the substance of this issue. Leaving open.

Issue #99, opened 2026-08-13
CI: supply-chain checks — network-egress canary job, pip-audit/OSV scanning, Dependency Review


alex-zwingli:
## Context

Builds on the CI foundation proposed in #95. Goal: catch a compromised or vulnerable dependency version *before* it merges, rather than after it's holding a user's Monarch token.

## Proposed checks

**1. Egress canary job.**
Install the locked dependencies, then import every module and run the full test suite with a socket-blocking audit hook active and **zero** hosts allowed. Any dependency that phones home at import time or during tests fails CI. This can literally reuse the runtime guard module proposed in #96, which keeps the two in sync — the CI canary proves the guard works, and the guard enforces in production what the canary verified at merge time.

**2. Known-bad screening.**
Run `pip-audit` (or `osv-scanner`) against `uv.lock` on every PR and on a weekly schedule. OSV carries malicious-package advisories (`MAL-*` IDs), not just CVEs, so this catches known-compromised releases, typosquats, and yanked-for-malware versions.

**3. GitHub Dependency Review action.**
On PRs that touch dependencies, blocks introduction of versions with known advisories. Zero-config once dependency graph is enabled.

**4. (Optional) behavioral analysis.**
Socket.dev's GitHub app (free for OSS) diffs dependency *behavior* across versions — new install scripts, new network/env/filesystem access, obfuscated code — which catches novel compromises that have no advisory yet.

**5. (Optional) harden-runner.**
StepSecurity harden-runner on the workflows themselves, so the CI runner's own egress is allowlisted and a compromised action or test dependency can't exfiltrate from CI.

## Priority

The egress canary (1) is the highest-value and most distinctive check — it converts "someone eventually notices the malware" into "the bump PR goes red." 2 and 3 are near-zero-cost table stakes.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code) after a source review of the repo.

robcerda:
Partially done in #132. pip-audit now runs against OSV and the PyPI advisory database on pull requests, pushes to main, and weekly, gated on a baseline so a newly published advisory fails the build while the known backlog stays quiet. Not done: the network egress canary, which is the hard part. Leaving this open for that.

PR #132, opened 2026-09-07
Add CI and supply chain workflows

### LINKED ISSUE #95: CI: add foundation workflow — tests, locked/hash-verified installs, SHA-pinned actions
## Context

The repo currently has no GitHub Actions workflows, but it already has a solid pytest suite (`tests/`) and a `uv.lock` with sha256 hashes for every dependency. That lockfile is doing no enforcement work today: nothing verifies that what gets installed (locally or in a future CI job) matches what was reviewed.

This issue is the foundation for a set of supply-chain hardening checks (filed separately) that assume CI exists.

## Proposal

Add `.github/workflows/ci.yml` that runs on PRs and pushes to `main`:

1. **Locked, hash-verified installs** — `uv sync --locked` fails the build if `uv.lock` drifts from `pyproject.toml`, and lockfile installs are hash-verified, so a PR can't smuggle in a substituted artifact.
2. **Run the test suite** — `uv run pytest` across the supported Python versions (3.12, 3.13 per `requires-python`).
3. **Least-privilege workflow** — top-level `permissions: contents: read`.
4. **Pin all actions by commit SHA** (not tags) so a compromised action tag can't alter the build.

## Why it matters for this project in particular

This server holds a long-lived Monarch Money session token. The threat model worth designing for is a compromised upstream release of `monarchmoneycommunity`/`gql`/`aiohttp` — the packages that hold the token in memory and talk to the network. Enforced lockfile installs are the first layer: they guarantee the artifact that was reviewed is the artifact that ships.

## Acceptance criteria

- CI runs on every PR; failing tests or a drifted lockfile block merge.
- All actions referenced by full commit SHA.
- README badge optional but nice.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code) after a source review of the repo.

robcerda:
Closes #95, and the tractable halves of #99 and #100. The repo had no CI at all.

## ci.yml
Tests on Python 3.12 and 3.13 via `uv sync --locked`, so CI cannot resolve a different dependency set than the lockfile pins. `uv lock --check` runs as its own job so drift reports separately. That is the check I most wanted: a bump recently changed pyproject without regenerating the lock, leaving the lock pinning the old version with hashes while the manifest asked for a newer one, which silently disabled hash verified installs.

No mypy or black gating. Neither passes cleanly today and a red build on day one is worse than no build.

## supply-chain.yml
pip-audit against both OSV and the PyPI advisory database, on pull requests, pushes to main, and weekly so advisories against untouched dependencies still surface.

The gate is a baseline delta. `.github/audit-baseline.txt` records the 35 advisories currently outstanding, each with its package and fix version. Anything not in that list fails the build. The standing backlog stays quiet, a newly published advisory turns the build red. Malicious package advisories are a separate always fail that a baseline entry cannot silence, though the MAL prefix alone is not sufficient coverage on its own: GitHub reviewed malware is not consistently mirrored into that namespace, so the canonical ctx compromise is a plain GHSA. The delta gate is what catches that.

There is no Dependency Review job. It was removed rather than fixed. The dependency graph is disabled here, and enabling it would not help much: GitHub's graph does not parse `uv.lock`, and requirements.txt and pyproject.toml carry only lower bounds, so there is nothing concrete for it to evaluate while every transitive change lands in the lock.

## Verification
Both workflows green. The audit run reports 78 locked entries audited against both services, 35 findings, all baselined, zero new.

Every action pinned to a full SHA, each re-resolved against its tag. The uv checksum matches the published sha256 for that release byte for byte, and is passed explicitly because setup-uv's built in table stops at uv 0.12.4 while this pins 0.12.10, so without it the binary was fetched unverified. No `${{ }}` in any run block, least privilege permissions, no pull_request_target, no SARIF so the workflow never needs security-events write, and persist-credentials false on every checkout.

Environment markers are stripped before auditing. pip-audit silently skips requirements whose marker is false on the runner without recording them as skipped, so the win32 pins that Windows users install were going unaudited. That is the difference between 75 and 78 entries.

Two rounds of adversarial review were applied. The material finding: audit reports were written into the checked out workspace, and pip-audit does not truncate its output on a fatal error, so a fork PR could commit a clean looking report, force the fatal with a hashless requirement in uv.lock, and the gate would report no advisories. Reproduced, then fixed by writing reports under RUNNER_TEMP.

## Known limits, not addressed here
- Nothing in this PR blocks a merge. main has no required status checks and no rulesets. The check names have to report once before they can be marked required, so that settings change has to follow this merge.
- setuptools and wheel are outside uv.lock entirely and fetched unpinned on every build.
- requirements.txt is a hand maintained parallel manifest that no check validates, and the README points users at it.
- A hand edited uv.lock pointing at an attacker published release with a matching hash passes every check here. That is #100, still open.

## Worth acting on separately
mcp 1.13.1 carries 4 advisories. Constraining to `mcp[cli]>=1.28.1,<2` resolves to 1.29.1 and adds only pyjwt, clearing those 4. It does not clear the starlette or python multipart findings, which stay at 0.47.3 and 0.0.20 respectively. Clearing those needs a broad upgrade that also moves mypy, pytest and starlette across major versions.

### DIFF
--- a/.github/audit-baseline.txt
+++ b/.github/audit-baseline.txt
@@ -5,9 +5,15 @@
 # Every entry is a deliberate, dated acknowledgement. Remove entries as
 # dependencies are upgraded. Do not add one without saying why.
 #
-# Seeded 2026-09-07 from the first CI run of this workflow.
-# Nearly all of these are transitive under mcp[cli] and clear by
-# upgrading mcp, which is tracked separately.
+# Seeded 2026-09-06 from the first CI run of this workflow.
+#
+# Do not assume an mcp bump clears this. It does not. Measured against the
+# committed lock: constraining to mcp[cli]>=1.28.1,<2 resolves to 1.29.1 and
+# clears exactly the three mcp entries. Every starlette and python-multipart
+# entry survives, because mcp depends on them with lower bounds only and uv
+# resolves minimally, so a transitive pin that already satisfies the new
+# constraint does not move. Clearing those needs an explicit
+# uv lock --upgrade-package starlette python-multipart, or a full upgrade.
 
 # black 25.1.0
 PYSEC-2026-2120  # fixed in 26.3.0
--- a/.github/audit-baseline.txt
+++ b/.github/audit-baseline.txt
@@ -1,0 +1,71 @@
+# Known advisories against the locked dependency set, accepted for now.
+# The audit fails on any advisory NOT listed here, so a newly published
+# one is a red build while this standing backlog stays quiet.
+#
+# Every entry is a deliberate, dated acknowledgement. Remove entries as
+# dependencies are upgraded. Do not add one without saying why.
+#
+# Seeded 2026-09-07 from the first CI run of this workflow.
+# Nearly all of these are transitive under mcp[cli] and clear by
+# upgrading mcp, which is tracked separately.
+
+# black 25.1.0
+PYSEC-2026-2120  # fixed in 26.3.0
+PYSEC-2026-2121  # fixed in 26.3.1
+
+# click 8.2.1
+PYSEC-2026-2132  # fixed in 8.3.3
+
+# cryptography 45.0.7
+GHSA-537c-gmf6-5ccf  # fixed in 48.0.1
+PYSEC-2026-2141  # fixed in 46.0.5
+PYSEC-2026-35  # fixed in 46.0.6
+PYSEC-2026-3552  # fixed in 50.0.0
+PYSEC-2026-3553  # fixed in 49.0.0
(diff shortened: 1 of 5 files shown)


robcerda:
Reviewed adversarially twice, signed off by a principal and a staff review, then independently validated with instructions to refute rather than confirm. That last round overturned two claims I had written into the files: that GitHub's dependency graph cannot parse uv.lock, and that upgrading mcp clears most of the advisory backlog. Both corrected in place, along with the one real fail open, where an empty export would have been reported as no known advisories.

PR #139, opened 2026-09-07
get_accounts: expose sync status and last-updated time

alexmv:
When an institution link breaks, the account keeps its last balance and stops importing transactions; the API raises no error for it. The GetAccounts query already returns the credential metadata that reveals this (updateRequired, disconnectedFromDataProviderAt, syncDisabled, displayLastUpdatedAt), but get_accounts dropped it, so the basic listing could not distinguish a healthy account from one frozen for a year without a second call to get_account_sync_health.

Add last_updated_at, is_manual and a sync block with a single state (ok / needs_reauth / disconnected / sync_disabled / manual) plus the raw fields. get_account_sync_health remains the per-connection view with staleness thresholds; this makes the listing itself trustworthy.

### DIFF
--- a/src/monarch_mcp_server/tools/accounts.py
+++ b/src/monarch_mcp_server/tools/accounts.py
@@ -28,9 +28,50 @@ class BalanceCorrections(RootModel[Dict[date, Decimal]]):
     """
 
 
+def _sync_status(account: dict) -> dict:
+    """Summarise an account's institution-connection state.
+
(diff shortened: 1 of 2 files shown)


robcerda:
Verified the premise: all seven fields are already selected by GetAccounts in monarchmoneycommunity 1.5.2, with the nesting matching how the code reads them, so this costs no extra API call. Purely additive to the output, and complementary to get_account_sync_health rather than overlapping it. Merging.

PR #141, opened 2026-09-11
ci: publish container images on GitHub releases

kquinsland01:
After the Docker/HTTP implementation in https://github.com/robcerda/monarch-mcp-server/pull/142 is merged, this PR adds GitHub Actions automation to build the root Dockerfile and publish a Linux AMD64 image to `ghcr.io/<owner>/<repository>` whenever a release is published.

### Merge order

1. Merge #142 for the Dockerfile, HTTP transport, tests, and deployment README.
2. Merge this PR for release publishing and manual builds.
3. Manually publish a release whose tag contains both changes.

This PR changes only `.github/workflows/ci_build-push-container.yaml` and `docs/container-publishing.md`. The Dockerfile, application code, and deployment README are supplied by #142, so the publishing change can be merged afterward without replacing them.

The workflow uses `GITHUB_TOKEN` with `packages: write`. Published images receive the release tag and a commit tag; stable releases also update `latest`. Manual `workflow_dispatch` remains available for the selected branch or tag and publishes branch/tag plus commit tags. Manual runs and prereleases do not automatically update `latest`.

### Validation

- Actionlint and `git diff --check` passed.
- `git merge-tree` confirms a clean merge after #142. The resulting tree adds only the workflow and publishing documentation, preserving #142's Dockerfile, HTTP code, tests, and README.
- The Docker image and HTTP/STDIO runtime checks are covered in #142.

GHCR publishing will be exercised when the workflow runs on GitHub. This PR remains a draft pending the Docker/HTTP work.

PR #142, opened 2026-09-11
feat: add Docker hosting with Streamable HTTP transport

kquinsland01:
This is a minimally invasive change to add streamable http support and basic Docker container image building.

See #141 for the github actions to build and push the container automatically.

The Docker image listens on `0.0.0.0:8000`; local launches keep STDIO as the default.
Transport, host, port, and allowed Host/Origin headers can be configured through CLI flags or environment variables.

The Astral uv/Python 3.12 image installs runtime dependencies from `uv.lock` and runs as a non-root user.

The README separates local installation, container deployment, and HTTP configuration, including authentication through a persistent session volume and reverse-proxy deployment.

All connected clients share one Monarch account; remote access requires an authenticated proxy or a private network with access controls and I did my best to call this out _twice_ so it's hard to ignore the implications.

### Merge order

1. Merge this PR to add the Docker image, HTTP transport, tests, and deployment documentation.
2. Merge https://github.com/robcerda/monarch-mcp-server/pull/141 to add release-triggered GHCR publishing and manual workflow dispatch on top of this Dockerfile.

### Validation

- All 400 tests pass, including HTTP initialization, tool discovery, a mocked tool call, configuration precedence, and Host/Origin validation.
- Built the Dockerfile with Podman and connected through a published HTTP port using the MCP SDK client: initialization, all 58 tools, and the expected response before Monarch authentication passed.
- Verified the Docker STDIO override with an MCP client.
- `git diff --check` passed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+# Send only the files needed to build the application, excluding local credentials.
+*
+!Dockerfile
+!uv.lock
+!pyproject.toml
+!README.md
+!LICENSE
+!login_setup.py
+!src/
+!src/**
+**/__pycache__
+**/*.py[cod]
+**/*.egg-info
+**/.env*
+**/*.pickle
+**/*.session
+**/.monarch-mcp-server
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+# Keep the uv version aligned with .github/workflows/ci.yml.
+FROM ghcr.io/astral-sh/uv:0.12.10-python3.12-trixie-slim
+
+ENV PYTHONDONTWRITEBYTECODE=1 \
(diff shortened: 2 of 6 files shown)


robcerda:
Reviewed the security posture closely. STDIO stays the default so existing setups are untouched, host defaults to 127.0.0.1 outside Docker, every docker run example publishes to loopback only, and the docs are accurate that Host and Origin checks stop DNS rebinding rather than authenticating callers. I verified that last point against a running server: a forged Host of localhost is accepted from any client, so the reverse proxy guidance is doing the real work and the README says so.

TransportSecuritySettings exists in the pinned mcp 1.13.1, so this works on the locked version. 400 tests pass including the negative Host cases. Merging.

PR #145, opened 2026-09-13
fix: include synced statements in recurring transaction reads

nixonthelobster:
## Problem

`get_recurring_transactions` delegates to the SDK's merchant-only recurring query. Synced credit-card and loan bills shown on Monarch's Recurring page are omitted, and the tool response does not expose their statement balances or due dates.

## Changes

- Query recurring items with `includeLiabilities` enabled by default; provide an explicit merchant-only opt-out.
- Preserve the existing list response and merchant fields while adding payment-account, liability-account, and statement identity.
- Return original statement amount, remaining balance, minimum payment, payment status, and due date independently. Preserve nulls and zeros rather than substituting forecast amounts.
- Add bounded page controls and an opt-in metadata envelope. Unknown totals remain null; full pages conservatively report possible truncation.
- Document synced bills versus forecasts, possible overlap between loan bills and merchant predictions, and pagination.

The tool performs reads only. It does not enable Bill Sync, refresh institutions, deduplicate obligations, calculate amounts owed, or change account data.

## Verification

- `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring .venv/bin/python -m pytest -q`: **424 passed** on Python 3.13.
- 24 focused regression cases using synthetic fixtures. The missing-bill regression fails against the previous implementation and passes with this patch.
- Read-only authenticated MCP smoke test verified schema discovery, liability inclusion, merchant-only opt-out, and equality of paginated versus single-page results. No account data or credentials are included in this PR.
- `git diff --check` passes.

Compatibility: legacy callers still receive a list, but now see synced bills when configured. `include_liabilities=False` requests merchant-only results; `include_metadata=True` exposes pagination metadata.


nixonthelobster:
cc my human @pjguzman

nixonthelobster:
Hi @robcerda, a quick follow-up on this synced-statements fix. The Python 3.12/3.13 tests, lockfile check, and dependency audit are passing, and GitHub currently reports no merge conflicts. Could you take a look when you have a chance? Happy to address any feedback or adjust the approach. Thanks!

robcerda:
Thank you for this! I really appreciate the work, especially the careful handling of nulls and zeros and the clear notes on where synced bills and merchant forecasts can overlap. Surfacing the bills Monarch shows on its own Recurring page is a real improvement. Merging now. I'll follow up separately so the default list response can't silently stop at 100 items.

PR #146, opened 2026-09-14
Fix Windows/cp1252 failures: explicit UTF-8 and two POSIX-only tests

cvanaver:
On a default Windows install (console code page cp1252) the setup script cannot complete and the suite reports 7 failures. Both come down to text encoding being left to the locale.

**Before:** `7 failed, 393 passed` · **After:** `400 passed, 1 skipped`

### 1. Session file fallback writes UTF-8 and reads locale encoding

`_write_secret_file` opens with `encoding="utf-8"`, but `_load_token_file` read it back with `Path.read_text()` and no encoding, which uses the locale's preferred encoding.

An ASCII blob survives that asymmetry, which is why it hasn't been noticed — the DPAPI-encrypted form is base64. A blob containing any non-ASCII character (a display name with an accent, say) either raises `UnicodeDecodeError` or decodes to a different string than was stored, so the credential comes back **silently corrupt rather than missing**.

One-line fix, plus a regression test that round-trips a non-ASCII blob. The test fails without the fix — I verified by reverting it.

### 2. `login_setup.py` dies on its own output

Every prompt and status line contains emoji. On cp1252 the first `print()` raises `UnicodeEncodeError` and the script exits **before the auth menu**, with an encoding traceback that says nothing about logging in — so a new Windows user can't authenticate at all and has no obvious way to find out why.

`stdout`/`stderr` are now reconfigured to UTF-8 with `errors="replace"`, so an un-renderable character can't be fatal again. Verified by running under a forced `PYTHONIOENCODING=cp1252`: previously an immediate traceback, now it reaches the menu.

### 3. Two file-fallback tests assume POSIX storage

Neither is a product bug — they fail on Windows against a correct implementation:

- `test_file_is_never_world_readable` asserts mode `0600`/`0700`. Windows `os.stat()` synthesises a mode that ignores the ACL, so the assertion can't express the property. Skipped there with the reason recorded; the Windows equivalent (DPAPI encryption, which makes the bytes unreadable to another local user whatever the ACL says) is already covered by the DPAPI tests.
- `test_write_is_atomic_via_replace` looks for the old token as plaintext mid-write. On Windows the blob is DPAPI-encrypted at rest, so plaintext is never there and the test failed while atomicity itself held. It now decrypts before asserting, checking the atomicity it's named for rather than the storage format.

If you'd rather keep the skip out of the suite, I'm happy to assert the ACL via `win32security` instead — it's more code and another dependency, so I went with the smaller change.

### Also fixed

Four `README.read_text()` calls in `tests/test_readme_tool_reference.py` and `tests/test_read_only.py` hit the same locale-encoding problem against a README containing non-cp1252 characters. Those four tests are what made the tool-table contract unverifiable on Windows.

Tested on Windows 11, Python 3.13, `uv`. No behaviour change on macOS/Linux — every change either adds an explicit encoding that matches what was already being written, or is Windows-conditional.


robcerda:
Both bugs confirmed. I reproduced the encoding asymmetry and my first attempt wrongly showed it round tripping, because json.dumps escapes non ASCII by default, which is exactly the reason you gave for it going unnoticed. Forcing real non ASCII bytes shows Jose coming back mangled. I also ran login_setup.py under a cp1252 console and it dies on the first banner line, so Windows users could not authenticate at all. Thanks for fixing the two POSIX assumptions in my tests as well.

